PR #1, opened 2025-04-03
feat: Genesis

tlgimenes:
This PR adds much of the code that we were doing in webdraw repo. The goal is to make it an open source ai agent creator


phdro88:
I'm starting the analysis of connections. First, I'll list the organizations and connections.

PR #70, opened 2025-04-12
Feat: Add Handoff view

jovenan:
Feat: Add Handoff view
Now we receive from api the list of tools, agents and core integrations

https://github.com/user-attachments/assets/db063d03-e07d-4176-ae77-1880ee9533b1

## Review Checklist
- [ ] No breaking changes

### DIFF
--- a/packages/sdk/src/hooks/agent.ts
+++ b/packages/sdk/src/hooks/agent.ts
@@ -176,7 +176,7 @@ export const useMessages = (agentId: string, threadId: string) => {
 
   const data = useSuspenseQuery({
     queryKey: getKeyFor(context, agentId, threadId),
-    queryFn: () => agentStub.query({ threadId }),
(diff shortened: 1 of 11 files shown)


packages/sdk/src/hooks/agent.ts
```diff
@@ -176,7 +176,7 @@ export const useMessages = (agentId: string, threadId: string) => {
 
   const data = useSuspenseQuery({
     queryKey: getKeyFor(context, agentId, threadId),
-    queryFn: () => agentStub.query(),
```
viktormarinho, line 179: why was this change needed? since the agent stub is already created for a threadId

jovenan: you are right, before this doesnt work for some reason, but now, this work. I removed

mcandeia (approved):

PR #73, opened 2025-04-14
Improving chat scroll

viniciustrr:
Changed the scroll behavior: now, when the chat is streaming, the user can scroll freely. However, if the user is focused at the bottom while the chat is streaming, the scroll will automatically follow the stream.


0xHericles:
Nice, we could have some arrow down icon when the user scrolls up, similar to chat gpt

viniciustrr:
Added icon to scroll down 

Captura de Tela 2025-04-15 às 15 24 31

0xHericles (approved):

PR #78, opened 2025-04-15
fix: delete button visible on mobile

viniciustrr:
Delete file attachment on the mobile and tablet are now possible. Desktop behaviour still the same

Captura de Tela 2025-04-15 às 11 36 41


viniciustrr:
I noticed there is a bug related to multiple selection of files:

1. When you add a file and delete it and try to add it again it wont work
2. When you have a file attached and try to add another one it will replace the existing one

Both fixed.

Captura de Tela 2025-04-15 às 11 50 24

PR #79, opened 2025-04-15
fix: handle url alert from stripe redirects

viktormarinho:
when redirected from stripe with some alert on the querystring, show it like this:

Captura de Tela 2025-04-15 às 14 09 26


jovenan (approved):

viktormarinho:
pedrofranca approved ;p

PR #106, opened 2025-04-21
chore: split agent

tlgimenes:
Split agents into a few components so we can compose them in different views


apps/web/src/components/agent/ChatHeader.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </h1>
+    </Container>
+  );
+};
+
+ChatHeader.Skeleton = () => {
+  return <div className="h-10 w-full" />;
+};
+
+ChatHeader.UI = ({ agentId }: Props) => {
```
lucis, line 53: What's the UI of a component?

lucis (approved):

apps/web/src/components/agent/ChatHeader.tsx
tlgimenes: it's the part that neither the error fallback or skeleton

PR #156, opened 2025-04-29
feat: login at api app

igorbrasileiro:
Move login from auth.deco.chat (from webdraw) to api app


apps/api/src/app.ts
```diff
@@ -96,6 +98,21 @@ app.all("/mcp", async (c: Context) => {
   }
 });
 
+// Login and auth routes
+Object.entries(loginRoutes).forEach(([route, honoApp]) => {
+  app.route(route, honoApp);
+});
+
+app.all("/api/user", (c: AppContext) => {
```
tlgimenes: can we use the PROFILES_GET tool call?

apps/web/src/main.tsx
```diff
@@ -19,6 +19,7 @@ import { Layout } from "./components/layout.tsx";
 import Login from "./components/login/index.tsx";
 import { ErrorBoundary, useError } from "./ErrorBoundary.tsx";
 import { trackException } from "./hooks/analytics.ts";
+import MagicLink from "./components/login/magicLink.tsx";
```
viktormarinho: suggestion: lazy import

apps/api/src/auth/index.ts
```diff
@@ -1,10 +1,49 @@
...
+export const ROUTES = {
+  ["/auth"]: appAuth,
+  ["/login"]: appLogin,
+} as const;
+
+const createDbAndHeadersForRequest = (ctx: AppContext) => {
+  const { SUPABASE_URL, SUPABASE_SERVER_TOKEN } = getEnv(ctx);
+  const request = ctx.req.raw;
+  const { headers, setCookie } = authSetCookie({ request });
+  const db = createSupabaseClient(
```
tlgimenes, line 26: cant we use the getUser function?

tlgimenes (approved):

apps/api/src/auth/index.ts
igorbrasileiro: You mean, refactor the function getUser to return the user and the database that handles cookies from request?

apps/api/src/app.ts
igorbrasileiro: Sure! I will remove change the UI for this

apps/web/src/main.tsx
igorbrasileiro: fixed:

apps/api/src/app.ts
igorbrasileiro: fixed

PR #200, opened 2025-05-07
Invite Member Flow

lucis:
## Also:
- Automatically return User and Workspace management MCPs
- Default integration connection type is HTTP

image


viktormarinho:
closing. @igorbrasileiro did it in another PR

PR #201, opened 2025-05-07
remove messages that got limit in the history

guitavano:
Currently, when a message with too many tokens is sent or received, the chat fails because it always tries to include the full thread history.

not sure about use the number 200.000

https://www.loom.com/share/ccb8f8f1b7374e1bad7177d6ccffe27a


viktormarinho:
some considerations:
- string length is not the same thing as llm tokens
- the server also includes lots of tokens on the context, like instruction on the agent configuration and tools, so maybe we're better off calculating it there. actually, i believe we already have it or almost, you should check it there on the actor. ask me if you need help on locating it.

PR #209, opened 2025-05-08
better chat code snippet rendering

viktormarinho:
added syntax highlighting with prism, a copy button and better ui.

Before:
Captura de Tela 2025-05-08 às 13 53 39

Now:
Captura de Tela 2025-05-08 às 13 52 07

### DIFF
--- a/apps/web/src/components/chat/Markdown.tsx
+++ b/apps/web/src/components/chat/Markdown.tsx
@@ -1,5 +1,5 @@
 import { marked } from "marked";
-import { memo, useMemo } from "react";
+import { memo, useMemo, useRef } from "react";
 import ReactMarkdown from "react-markdown";
 import rehypeRaw from "rehype-raw";
 import remarkGfm from "remark-gfm";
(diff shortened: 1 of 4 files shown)


deno.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "lib": ["deno.window", "dom", "esnext", "deno.ns", "deno.unstable"],
+    "types": ["@cloudflare/workers-types"],
+    "jsx": "react-jsx",
+    "jsxImportSource": "react"
+  }
+}
```
igorbrasileiro: Only use package.json

viktormarinho: im seeing now that this is in package.json already. but for some reason only works in deno.json. i will remove it from this pr from now since it is unrelated actually

apps/web/src/components/chat/Markdown.tsx
```diff
@@ -59,19 +59,81 @@ const MemoizedMarkdownBlock = memo(
...
+function CodeBlock(
+  { language, content }: { language: string; content: string },
+) {
+  const [copied, setCopied] = useState(false);
+
+  function handleCopy() {
+    navigator.clipboard.writeText(content);
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1200);
+  }
```
igorbrasileiro:
```suggestion
  const  handleCopy = async () =>  {
    await navigator.clipboard.writeText(content);
    setCopied(true);
    setTimeout(() => setCopied(false), 1200);
  }
```

igorbrasileiro: Also this is buggy, 
1. click to copy;
2. after some time, between the first click and 1200ms window, click 

the copy values will be false before the 1200ms from second click because the timeout of first.

igorbrasileiro: So you could do this:

```ts
const timeoutRef = useRef(null);

const handleCopy = async () => { 
  ...
  clearTimeout(timeoutRef.current);
  timeoutRef.current = setTimeout(() => setCopied(false), 1200)
}

viktormarinho: done

igorbrasileiro (approved):

igorbrasileiro:
Just a quick hint, lazy load this component

PR #240, opened 2025-05-14
feat: auto model

viktormarinho:
auto model v1 - without an algorithm backing it up. the auto model is hardcoded on the server.

seems already useful tho, because users can keep this setting and we can change server-side the best general model at the moment.

image

### DIFF
--- a/apps/web/src/components/chat/ModelSelector.tsx
+++ b/apps/web/src/components/chat/ModelSelector.tsx
@@ -12,7 +12,7 @@ import {
   ResponsiveSelectValue,
 } from "@deco/ui/components/responsive-select.tsx";
 import { cn } from "@deco/ui/lib/utils.ts";
-import { DEFAULT_REASONING_MODEL, MODELS } from "@deco/sdk";
+import { DEFAULT_MODEL, MODELS } from "@deco/sdk";
 import { useState } from "react";
 
 const mapLegacyModelId = (modelId: string): string => {
@@ -129,7 +129,7 @@ interface ModelSelectorProps {
 }
 
 export function ModelSelector({
-  model = DEFAULT_REASONING_MODEL,
+  model = DEFAULT_MODEL,
   onModelChange,
   variant = "borderless",
 }: ModelSelectorProps) {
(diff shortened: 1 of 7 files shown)


tlgimenes (approved):

packages/sdk/src/constants.ts
```diff
@@ -73,6 +74,14 @@ type Capability =
   | "web-search";
 
 export const MODELS: Model[] = [
+  {
+    id: "auto",
+    name: "Auto",
```
igorbrasileiro, line 79: Why not name it`mixed` ?

packages/sdk/src/constants.ts
```diff
@@ -73,6 +74,14 @@ type Capability =
   | "web-search";
 
 export const MODELS: Model[] = [
+  {
+    id: "auto",
```
igorbrasileiro:
```suggestion
    id: DEFAULT_MODEL,
```

apps/web/src/components/chat/ModelSelector.tsx
```diff
@@ -45,6 +45,12 @@ const CAPABILITY_CONFIGS = {
     text: "text-amber-700",
     label: "Can search the web to answer questions",
   },
+  "mixed": {
```
mcandeia:
```suggestion
  mixed: {
```

packages/ai/src/agent.ts
```diff
@@ -3,7 +3,7 @@ import { createOpenAI } from "@ai-sdk/openai";
 import type { JSONSchema7 } from "@ai-sdk/provider";
 import type { ActorState, InvokeMiddlewareOptions } from "@deco/actors";
 import { Actor } from "@deco/actors";
-import { WELL_KNOWN_AGENTS } from "@deco/sdk";
+import { DEFAULT_MODEL, WELL_KNOWN_AGENTS } from "@deco/sdk";
```
mcandeia, line 6: why not DEFAULT_REASONING_MODEL?

viktormarinho: misleading as reasoning is often seen (in our own codebase even) as a characteristic of a model, even though the agent can work with models that don't perform "reasoning" or we can disable it

packages/sdk/src/constants.ts
viktormarinho: uhh i prefer auto

PR #248, opened 2025-05-14
feat: Publish agent 

tlgimenes:
https://www.loom.com/share/d6380661a2234387b8fc3b20f5ba9cd2?sid=6c9da2cb-dfcb-47e3-83d5-bfcb2006209f


lucis (approved):
Code looks good

Can't infer much from it about how the UX is like

PR #297, opened 2025-05-20
feat: auth on MCP tools

igorbrasileiro:
## Review Checklist
- [ ] No breaking changes


packages/sdk/src/auth/policy.ts
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/**
+ * PolicyClient - Singleton class for managing policy access
+ */
+export class PolicyClient {
+  private static instance: PolicyClient | null = null;
+  private db: Client | null = null;
+  private userPolicyCache: LRUCache<string, Policy[]>;
+  private teamRolesCache: LRUCache<number, Role[]>;
```
mcandeia: can we use `@deco/sdk/cache` instead? so we can cache across workers

mcandeia: a good approach would be:

1. cache for 1min~5min
2. delete before update (so we can keep it strongly consistent)

packages/sdk/src/mcp/agents/api.ts
```diff
@@ -52,19 +52,16 @@ export const listAgents = createApiHandler({
   name: "AGENTS_LIST",
   description: "List all agents",
   schema: z.object({}),
+  async canAccess(_, c) {
+    return await canAccessWorkspaceResource(this.name, c);
+  },
```
mcandeia: where `this.name` comes from? looks a bit implicit

perhaps you can set a default can access when creating an createApiHandler?

mcandeia (changes requested):

packages/sdk/src/mcp/agents/api.ts
igorbrasileiro: this.name comes from the APIHandlerDefinition. IMO, this.name is easy to get that it comes from the name defintion from the object.

packages/sdk/src/auth/policy.ts
igorbrasileiro: The code is deleting before update, at: https://github.com/deco-cx/chat/blob/6c09819964c1aea68193308c269a911d8af4fc5c/packages/sdk/src/auth/policy.ts#L309 or here https://github.com/deco-cx/chat/blob/6c09819964c1aea68193308c269a911d8af4fc5c/packages/sdk/src/auth/policy.ts#L179-L188

apps/api/package.json
```diff
@@ -20,6 +19,7 @@
     "ai": "^4.3.13",
     "cloudflare": "^4.2.0",
     "hono": "^4.7.7",
+    "lru-cache": "^11.1.0",
```
viktormarinho:
```suggestion
```

packages/sdk/src/mcp/agents/api.ts
igorbrasileiro: fixed, this

packages/sdk/src/auth/policy.ts
igorbrasileiro: done

viktormarinho (approved):

PR #314, opened 2025-05-21
feat: add reject invite

igorbrasileiro:

with invite
image

without invite
image

team list with invite
image

personal invites list
image


apps/web/src/components/sidebar/InvitesLink.tsx
```diff
@@ -0,0 +1,11 @@
+import { useInvites } from "@deco/sdk/hooks";
```
igorbrasileiro: rename to InvitesCount

PR #321, opened 2025-05-22
Fix local calls for non-authorized integrations

mcandeia:


packages/sdk/src/auth/user.ts
```diff
@@ -37,12 +41,15 @@ export async function getUserBySupabaseCookie(
...
+  const [{ data: _user }, jwt] = await Promise.all([
+    supabase.auth.getUser(
+      accessToken,
+    ),
+    jwtIssuer.verify(sessionToken),
+  ]);
   const user = _user?.user;
   if (!user) {
-    return undefined;
+    return jwt;
```
igorbrasileiro, line 52: What happens if jwtIssuer.verify fails? I mean, the sessionTOken isn't verified.

packages/sdk/src/mcp/assertions.ts
```diff
@@ -117,7 +129,8 @@ export const assertUserHasAccessToWorkspace = async (
 
   // TODO (@gimenes): remove this hard coded access for @deco.cx by allowing
   // deco.cx users to enter in any workspace
-  const isAdmin = user?.email?.endsWith("@deco.cx");
+  const isAdmin = user && typeof user === "object" && "email" in user &&
+    typeof user.email === "string" && user?.email?.endsWith("@deco.cx");
```
igorbrasileiro: Revert this?

packages/sdk/src/auth/user.ts
mcandeia: returns undefined, so same as prev behavior

PR #347, opened 2025-05-23
Feat: Add Edit trigger

jovenan:
Feat: Add Edit trigger

## Review Checklist
- [ ] No breaking changes


jovenan:
@claude review

packages/sdk/src/mcp/triggers/api.ts
```diff
@@ -211,6 +211,80 @@ export const createTrigger = createTool({
   },
 });
 
+export const updateTrigger = createTool({
```
mcandeia, line 258: When I have create/update methods I really expect to see the same underlying implementation and changing only insert to update or even better use upsert but validation/etc should be exactly the same. So I'd like to see something like 

handler: upsertTrigger

which is the same function as create, if possible

mcandeia (changes requested):

packages/sdk/src/mcp/triggers/api.ts
jovenan: for sure

PR #412, opened 2025-05-27
Feat: Add proxy to Assets R2 on worker

jovenan:
Feat: Add proxy to Assets R2 on worker
Now, you can have a static URL from R2 file.

Example of URL:
- https://assets.deco.chat/public/assets_integrations_icon-1e6a96f0-be46-42da-aab7-61df1b3f8cb5.png

## Review Checklist
- [ ] No breaking changes


apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+export const app = new Hono<AppEnv>();
+
+app.use(withContextMiddleware);
+
+app.use(
+  "/:root/:slug/*",
+  cache({
+    cacheName: "assets",
+    cacheControl: "public, max-age=31536000", // 1 year cache
+    wait: true, // Required for Deno environment
```
mcandeia: not always is deno, right?


import { getRuntimeKey } from "hono/adapter";
getRuntimeKey() === "deno"

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+  })
+);
+
+app.get("/:root/:slug/*", async (c) => {
```
mcandeia: why you need to split into `/:root/:slug` instead of just `/:workspace/*` ?

Also you can capture wildcard param:

`/:workspace/:path{.+}`

so you can access c.req.param('workspace') and c.req.param('path')

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    if (!c.req.param("root") || !c.req.param("slug")) {
+      throw new HTTPException(400, { message: "Workspace not found" });
+    }
```
mcandeia: this is not possible, you can remove. if this could be possible the routing is not working

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const workspace = `/${c.req.param("root")}/${c.req.param("slug")}/`;
+
+    const prefixIndex = c.req.path.indexOf(workspace);
+    const imagePath = c.req.path.substring(prefixIndex + workspace.length);
+
+    if (!imagePath) {
+      throw new HTTPException(400, { message: "Image path is required" });
+    }
```
mcandeia: there should always be a path, I think you don't need to check but not sure

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (!imagePath.startsWith("public/")) {
+      throw new HTTPException(403, { message: "Asset is not public" });
+    }
```
mcandeia: how about concat "public/${path}" so you don't need to validate?

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const bucketName = getWorkspaceBucketName(workspace);
+
+    const getCommand = new GetObjectCommand({
+      Bucket: bucketName,
+      Key: imagePath,
+    });
+
+    // deno-lint-ignore no-explicit-any
+    const response = await (c.var as any).s3.send(getCommand) as GetObjectCommandOutput;
```
mcandeia: c.var.s3 is not well typed? you need to check if it is properly defined at `AppEnv` it should be typed

apps/api/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (!response.Body) {
+      throw new HTTPException(404, { message: "Image not found" });
```
mcandeia: Image? assets can be other types rather than image, right?

packages/sdk/src/storage/s3/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+      .replace(/^-|-$/g, "")
+  }`;
+}
+
+/**
+ * Determines Content-Type based on file extension
+ * @param path File path
+ * @returns MIME type string
+ */
+export function getContentTypeFromPath(path: string): string {
```
mcandeia: you can use this too: https://jsr.io/@std/media-types

apps/api/src/assets.ts
jovenan: done

apps/api/src/assets.ts
jovenan: I changed this by using a public bucket

apps/api/src/assets.ts
jovenan: done

apps/api/src/assets.ts
jovenan: yes, done

apps/api/src/assets.ts
jovenan: I changed this to use a public bucket, so its not necessary

apps/api/src/assets.ts
jovenan: this is my lint error, I fixed and removed this. The s3 is already typed

apps/api/src/assets.ts
jovenan: yes

packages/sdk/src/storage/s3/utils.ts
jovenan: nice! Done

packages/sdk/src/mcp/fs/api.ts
```diff
@@ -196,17 +133,15 @@ export const deleteFile = createTool({
   inputSchema: z.object({ path: z.string() }),
   canAccess: canAccessWorkspaceResource,
   handler: async ({ path }, c) => {
-    const bucketName = getWorkspaceBucketName(c);
+    const bucketName = getWorkspaceBucketName(c.workspace!.value);
```
mcandeia:
```suggestion
    assertHasWorkspace(c)
    const bucketName = getWorkspaceBucketName(c.workspace!.value);
```

packages/sdk/src/mcp/fs/api.ts
```diff
@@ -168,20 +106,19 @@ export const writeFile = createTool({
   }),
   canAccess: canAccessWorkspaceResource,
   handler: async ({ path, expiresIn = 60, contentType, metadata }, c) => {
-    const bucketName = getWorkspaceBucketName(c);
+    const bucketName = getWorkspaceBucketName(c.workspace!.value);
```
mcandeia: assertHasWorkspace(c)

packages/sdk/src/mcp/fs/api.ts
```diff
@@ -3,75 +3,16 @@ import {
   GetObjectCommand,
```
mcandeia, line 3: use assertHasWorkspace instead of "!"

mcandeia (approved):

PR #438, opened 2025-05-29
feat: update design system + enforce with lint rule

viktormarinho:
- created a lint rule that warns when the user uses a token inside the JSX className attribute with a tailwind class arbitrary color.
- fix all lint warnings, so we now use only design system tokens
- update our design system


viktormarinho:
sidebar is done, need to check individual views using things like "slate-100"...

PR #475, opened 2025-06-04
chore: Access

tlgimenes:
This PR improves resource access by removing canAccess function and moving assertion logic into the handler function 

Also, a new helper function is added. You need to call ctx.requestAccess.grant() to ensure you have remembered to handle authorization in the tool

### DIFF
--- a/packages/sdk/src/mcp/wallet/api.ts
+++ b/packages/sdk/src/mcp/wallet/api.ts
@@ -176,8 +176,7 @@ export const getThreadsUsage = createTool({
   handler: async ({ range }, c) => {
     assertHasWorkspace(c);
 
-    await assertWorkspaceResourceAccess(c.tool.name, c)
-      .then(() => c.resourceAccess.grant());
+    await assertWorkspaceResourceAccess(c.tool.name, c);
 
     const wallet = getWalletClient(c);
 
@@ -199,8 +198,7 @@ export const getAgentsUsage = createTool({
   handler: async ({ range }, c) => {
     assertHasWorkspace(c);
 
-    await assertWorkspaceResourceAccess(c.tool.name, c)
-      .then(() => c.resourceAccess.grant());
+    await assertWorkspaceResourceAccess(c.tool.name, c);
 
     const wallet = getWalletClient(c);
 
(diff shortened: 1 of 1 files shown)


packages/sdk/src/mcp/agents/api.ts
```diff
@@ -202,9 +199,12 @@ export const deleteAgent = createTool({
   name: "AGENTS_DELETE",
   description: "Delete an agent by id",
   inputSchema: z.object({ id: z.string() }),
-  canAccess: canAccessWorkspaceResource,
```
igorbrasileiro, line 205: Agent should be deleted always by the creator, isn't?

packages/sdk/src/mcp/fs/api.ts
```diff
@@ -105,13 +107,15 @@ export const readFile = createTool({
...
-  canAccess: canAccessWorkspaceResource,
-  handler: async ({ path, expiresIn = 60 }, c) => {
+  handler: async ({ path, expiresIn = 60 }, c, { name }) => {
     const bucketName = getWorkspaceBucketName(c);
 
     assertHasWorkspace(c);
     await ensureBucketExists(c, bucketName);
 
+    await assertWorkspaceResourceAccess(name, { path, expiresIn }, c)
+      .then(() => c.resourceAccess.grant());
```
igorbrasileiro: Move the `.then( () => c.resourceAccess.grant())` to assertWorkspaceResourceAccess

packages/sdk/src/mcp/agents/api.ts
```diff
@@ -92,46 +94,36 @@ export const getAgent = createTool({
...
+          .eq("id", id)
+          .single(),
+    ]);
 
     if ((error && error.code == NO_DATA_ERROR) || !data) {
       throw new NotFoundError(id);
     }
 
+    if (data.visibility !== "PUBLIC" && !canAccess) {
+      throw new NotFoundError(`Agent ${id} not found`);
```
igorbrasileiro: shouldn't this return forbidden, should?. And it's not returning

packages/sdk/src/mcp/agents/api.ts
tlgimenes: i'm not sure. If they have access yes, if not,no

igorbrasileiro (approved):

PR #490, opened 2025-06-09
fix: update Vite build targets to support top-level await

pedrofrxncx:
Resolves build error with @jsr/deco__actors package - Updated targets to ES2022 and modern browsers (Chrome 89+, Firefox 89+, Safari 15+) 

image

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -15,7 +15,7 @@
     "@deco/sdk": "^0.1.1",
     "@deco/ui": "^1.0.0",
     "@gsap/react": "^2.1.2",
-    "@hookform/resolvers": "^5.0.1",
+    "@hookform/resolvers": "^3.3.4",
     "@tanstack/react-query": "^5.72.0",
     "@tiptap/extension-placeholder": "^2.11.7",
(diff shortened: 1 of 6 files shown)


apps/web/package.json
```diff
@@ -15,7 +15,7 @@
     "@deco/sdk": "^0.1.1",
     "@deco/ui": "^1.0.0",
     "@gsap/react": "^2.1.2",
-    "@hookform/resolvers": "^5.0.1",
+    "@hookform/resolvers": "^3.3.4",
```
igorbrasileiro, line 18: Don't downgrade

apps/web/vite.config.ts
```diff
@@ -6,4 +6,15 @@ import { defineConfig, type PluginOption } from "vite";
...
+    target: ["es2022", "chrome89", "firefox89", "safari15"],
+  },
+  esbuild: {
+    target: "es2022",
+  },
+  optimizeDeps: {
+    esbuildOptions: {
+      target: "es2022",
+    },
+  },
```
igorbrasileiro, line 19: Don't change this because we want to support old browsers

PR #498, opened 2025-06-10
fix(connections): newly created custom connections with single instance appearing as unknown

viktormarinho:
@claude review


viktormarinho:
@claude review

Issue #512, opened 2025-06-12
[Bug]: "Please allow popups" when connecting to oAuth MCP Server


lucis:
![Image](https://github.com/user-attachments/assets/fa3042b1-dcec-4cca-82b2-3ca49fde6456)

lucis:
Seems to be intermitent 

cc @viktormarinho

PR #513, opened 2025-06-12
Oauth fallback

lucis:
Prompt:
```md
There's a problem with this logic that sometimes it will work and sometimes it will show this alert. But the link creation is not working

I want to do something different. Remove the alert and the <a href creation as fallback and, instead, crea te a <Dialog with Finish Authentication Flow and inside the Dialog there's a link for the user to click.

After finish, do the same for @select-connection-dialog.tsx that has a similar flow
```

image


lucis:
Fixes #512

viktormarinho (approved):
nice

PR #525, opened 2025-06-13
fix: initial agent xp

viktormarinho:
- fix tools starting as open
- update new agent template model, token limit and starting tools


viktormarinho:
Captura de Tela 2025-06-13 às 11 17 10

cannot merge this atm... the team agent cannot update itself

PR #573, opened 2025-06-18
fix(memory): implement message processing workaround for tool-call a…

pedrofrxncx:
…nd tool-result messages

## Review Checklist
- [ ] No breaking changes


pedrofrxncx:
there is always a chance of introducing a new bug, so feel free to revert; this is a cheap fix

PR #587, opened 2025-06-19
Search integration filter tools instead

lucis:
**Prompt:**
```
I want to change the search mechanism here

It shouldn't search based on Integration's name, but on the integration's tools. And it should filter the tool list based on the search.

The goal is, for integrations with lots of tools, I can search by tool name to better activate/deactivate it

Plan and implement this change
```

**Before**
before-tools

**After:**
after-tools

### DIFF
--- a/apps/web/src/components/toolsets/selector.tsx
+++ b/apps/web/src/components/toolsets/selector.tsx
@@ -85,20 +85,22 @@ export function IntegrationListItem({
   // Helper function to check if a tool matches the search term
   const toolMatchesSearch = (tool: MCPTool, searchTerm: string) => {
     if (!searchTerm) return true;
-    
+
     const lowerSearchTerm = searchTerm.toLowerCase();
     const toolNameFormatted = beautifyToolName(tool.name).toLowerCase();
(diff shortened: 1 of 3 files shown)


pedrofrxncx (approved):

apps/web/src/components/toolsets/selector.tsx
```diff
@@ -80,6 +82,34 @@ export function IntegrationListItem({
   const isAll = enabledCount === total && total > 0;
   const isEmpty = !isLoading && allTools.length === 0;
 
+  // Helper function to check if a tool matches the search term
+  const toolMatchesSearch = (tool: MCPTool, searchTerm: string) => {
+    if (!searchTerm) return true;
+
+    const lowerSearchTerm = searchTerm.toLowerCase();
```
igorbrasileiro: Instead of lowerCase for every tool, i'd lowercase outside this function doing once.

Suggestion:

```ts
const searchText = useDefferedValue(_searchText.toLowerCase());
```

lucis: image

PR #669, opened 2025-06-27
Add output schemas to Agents and Integration apps

lucis:


packages/sdk/src/mcp/agents/api.ts
```diff
@@ -79,6 +79,7 @@ export const listAgents = createTool({
   name: "AGENTS_LIST",
   description: "List all agents",
   inputSchema: z.object({}),
+  outputSchema: z.array(AgentSchema),
```
tlgimenes, line 82: I think output schema can not be an array. Take a look at:

https://github.com/modelcontextprotocol/typescript-sdk/pull/528

Issue #677, opened 2025-06-28
Show last used workspace on top of Switch team selector


lucis:
- Create a LAST_USED_WORKSPACES = [] in localstorage 
- Every time a new workspace is loaded (can be checked in the Team Selector componente), check if workspace slug is already in the localstorage var. If not, add it 
- In the workspace list from Switch Team, sort by last used first (if they're in the list, as users can leave workspaces)

PR #678, opened 2025-06-28
Show most recent workspace first in Switcher

lucis:
Generated by Cursor based on #677


tasks/sort-workspaces.md
```diff
@@ -0,0 +1,34 @@
+Title: Show last used workspace on top of Switch team selector
```
lucis, line 1: What do you think about keeping those in source control?

PR #689, opened 2025-06-30
chore bundle apps locally

tlgimenes:
Uses local wrangler install to bundle at the dist folder. Then uploads the resulting js


viktormarinho:
@claude review

PR #701, opened 2025-07-01
chore: drop channels

igorbrasileiro:
## Review Checklist
- [ ] No breaking changes

### DIFF
--- a/apps/web/src/components/settings/advanced.tsx
+++ b/apps/web/src/components/settings/advanced.tsx
@@ -413,7 +413,6 @@ function AdvancedTab() {
             )}
 
             <div className="border-t pt-6">
-              <Channels />
             </div>
           </form>
         </div>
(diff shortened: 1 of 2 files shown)


apps/web/src/components/dock/index.tsx
```diff
@@ -51,6 +51,17 @@ export const useDock = () => {
 const adapter =
   <T extends object>(Component: ComponentType<T>) =>
   (props: IDockviewPanelProps<T>) => {
+    const [isActive, setIsActive] = useState(props.api.isActive);
+
+    useEffect(() => {
+      const { dispose } = props.api.onDidActiveChange((e) => {
+        console.log(props.api.id);
+        setIsActive(e.isActive);
```
tlgimenes: only set when going from false -> true. 
true -> false harms performance

igorbrasileiro: Good catch

PR #714, opened 2025-07-02
fix: add vite-plugin-pwa to solve dynamic import error

igorbrasileiro:
Use vite pwa plugin to solve dynamic import error cross deployment

### DIFF
--- a/apps/web/src/main.tsx
+++ b/apps/web/src/main.tsx
@@ -172,6 +172,24 @@ function ErrorFallback() {
     globalThis.location.href = `/login?next=${next}`;
   }, [isUnauthorized, pathname]);
 
+  useEffect(() => {
+    if (!error) return;
+    const wellKnownErrors = [
+      isUnauthorized,
+      error instanceof ForbiddenError,
+      error instanceof NotFoundError,
+    ]
+    if (wellKnownErrors.some(Boolean)) {
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


apps/web/src/main.tsx
```diff
@@ -172,6 +172,24 @@ function ErrorFallback() {
     globalThis.location.href = `/login?next=${next}`;
   }, [isUnauthorized, pathname]);
 
+  useEffect(() => {
```
tlgimenes: eu acho bom deixar o botao de retry la. Isso faz a gente ver o erro e ser reportado. 
Pode tambem ate dar um refresh loop

apps/web/vite.config.ts
```diff
@@ -1,10 +1,25 @@
...
 import { defineConfig, type PluginOption } from "vite";
+import { VitePWA } from 'vite-plugin-pwa';
 
 // https://vite.dev/config/
 export default defineConfig({
-  plugins: [react() as PluginOption[], tailwindcss() as PluginOption[]],
+  plugins: [
+    react() as PluginOption[], 
+    tailwindcss() as PluginOption[],
+    VitePWA({
```
tlgimenes, line 11: legal, vamos testar!

apps/web/src/main.tsx
igorbrasileiro: Good catch, i will drop the automatic reload

PR #719, opened 2025-07-03
Install all deployed apps as MCPs.

mcandeia:


mcandeia:
approved offline by @tlgimenes

PR #726, opened 2025-07-03
Table and avatar UI fixes

muridias:
## Review Checklist
- [ ] No breaking changes

### DIFF
--- a/apps/web/src/components/common/avatar/index.tsx
+++ b/apps/web/src/components/common/avatar/index.tsx
@@ -12,16 +12,16 @@ import { isFilePath } from "../../../utils/path.ts";
 
 // Predefined color palette for avatar backgrounds
 const AVATAR_COLORS = [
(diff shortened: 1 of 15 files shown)


apps/web/src/components/common/avatar/index.tsx
```diff
@@ -95,6 +95,47 @@ export function Avatar({
     return AVATAR_COLORS[0];
   }, [fallback]);
 
+  // Determine font size based on avatar size
+  const fontSize = useMemo(() => {
```
viktormarinho, line 99: these `useMemo`s for getting the borderRadius and fontSize are not cool for maintenance; it is not obvious the classes im ending up applying from outside the component IMO.

It should be done using variants with the `cva` function. take as reference another component of the `ui` package that uses that function, for example the `Button`.

After that, make every avatar pass the "size" variant instead of thoe classes inside classname

PR #710, opened 2025-07-02
feat: thread title gen, ai gateway mcp + fix wallet

lucis:
This PR introduces two major features: a stateless AI generation tool and automatic, AI-powered title generation for new threads. It also includes a refactor to remove the `uniqueByAgentId` functionality and a fix for handling negative values in the wallet system.

### Features

* **Stateless `AI_GENERATE` Tool**:
    * Adds a new `AI_GENERATE` tool within a new `AI` group, allowing for direct, stateless calls to LLMs without requiring an agent.
    * Integrates with the wallet system to charge for usage, introducing a new `LLMGeneration` transaction type.
    * Includes new database policies (`call_ai_gateway`) to manage access to this new capability.

* **Automatic AI Thread Title Generation**:
    * When a new chat thread is created, it now automatically generates a concise title based on the content of the first message using the new `AI_GENERATE` tool.
    * The UI optimistically displays a temporary title and then updates it with a streaming animation as the AI-generated title is received.

### Refactor

* **Remove `uniqueByAgentId`**:
    * The `uniqueByAgentId` option has been removed from the thread fetching logic on both the client-side (`useThreads` hook) and the backend API (`listThreads` tool).

### Fix

* **`MicroDollar` Negative Value Handling**:
    * The `MicroDollar` class in the wallet package has been updated to correctly handle, format, and perform arithmetic with negative values.
    * Adds a comprehensive test suite to validate the new `MicroDollar` functionality.

PR #727, opened 2025-07-04
Adding rules for Data Flow and Wallet

lucis:
To improve our AIX (AI development experience), I'm adding two new [Cursor rules](https://docs.cursor.com/context/rules):

- **Data Flow** describing tools, MCP servers, React Query.
- **Wallet** talking about our credit system.

I asked Cursor to create both of them based on @viktormarinho's work on #710

## Agent Requestes Rule

One interesting aspect is that I used this type of rule which is not included in every request but **the agent automatically requests it** if the current task is related to a Description field. This seems interesting.

image

(Check the Description)

Cursor 2025-07-03 21 27 46

And it works!

## Demo

https://github.com/user-attachments/assets/780068a9-87c0-4bde-8ff0-9cca9d46fdfa

### DIFF
--- a/.cursor/rules/data-flow.mdc
+++ b/.cursor/rules/data-flow.mdc
@@ -1,0 +1,337 @@
+---
+description: When working on anything related to dataflow, creating new tools, editing some functionality that envolves backend and frontend. Mentions Tools, Authorization and so on
+globs: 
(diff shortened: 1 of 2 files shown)


igorbrasileiro (approved):
LGTM

PR #811, opened 2025-07-15
Add default Open command for deco CLI login command

jvitormelo:
### When running deco login on Linux I got this error 

image

To fix it, I just added a default value to use if no env is declared.

I also formatted with the deno, and it made it look like I changed the entire file. I think I'm missing something. If you know what I did wrong, I can try to run again and format correctly

I was able to work on Linux, and will work on Mac (default command is open), but I don't have windows to test the default value

### DIFF
--- a/packages/cli/src/login.ts
+++ b/packages/cli/src/login.ts
@@ -18,7 +18,7 @@ export const loginCommand = async () => {
             illumos: "xdg-open",
             android: "xdg-open",
             freebsd: "xdg-open",
-            netbsd: "xdg-open", 
+            netbsd: "xdg-open",
             windows: "start",
             darwin: "open",
             aix: "open",


mcandeia:
LGTM @jvitormelo,

In order to release a new version the https://github.com/deco-cx/chat/blob/main/packages/cli/deno.json#L3 must be bumped, that's the only missing part, once done I can merge

packages/cli/src/login.ts
```diff
@@ -7,60 +7,78 @@ export const loginCommand = async () => {
...
+
+          switch (Deno.build.os) {
+            case "linux":
+              return "xdg-open";
+            case "windows":
+              return "start";
+            default:
+              return "open";
+          }
+        };
```
mcandeia: nit:

Suggestion only:
```suggestion
        const command = Deno.env.get("BROWSER") ?? { linux: "xdg-open", windows: "start" }[Deno.build.os] ?? "open";
```

jvitormelo: image

I will need to cast as Record<string, string> because typescript want to implement for all os
image

What do you think?

mcandeia: Actually I think it might be good to have all of those, why not, right?

Ask LLM to fulfill the remaining options, they are not many:

image

jvitormelo: Done!

pedrofrxncx (approved):

PR #841, opened 2025-07-18
Adds tool description as jsDoc

mcandeia:


packages/cli/src/typings.ts
```diff
@@ -307,8 +313,30 @@ ${tsTypes}
     props.filter((p) => p !== null).map(([propName, tools]) => {
       return `${propName}: Mcp<{
         ${
-        tools.map(([toolName, inputName, outputName]) => {
-          return `
+        tools.map(([toolName, inputName, outputName, description]) => {
+          // Sanitize description for safe use in JSDoc block comments
+          const formatDescription = (desc: string | undefined) => {
```
igorbrasileiro: Does it make sense to put it out of the map function?

mcandeia: fixed!

PR #844, opened 2025-07-18
chore: migrate formatter

viktormarinho:


.github/workflows/test.yml
```diff
@@ -24,12 +24,12 @@ jobs:
       - name: Install dependencies
         run: deno install
 
+      - name: Run format
+        run: deno run fmt:check
```
igorbrasileiro, line 28: remove deno and use npm or bun or any other package manager?

PR #868, opened 2025-07-22
fix(cli): `deco deploy` envVars reading

viktormarinho:


viktormarinho:
@claude review

packages/cli/src/wrangler.ts
```diff
@@ -9,18 +9,30 @@ import { readSession } from "./session.ts";
 
 const envFile = ".dev.vars";
 
-export async function getCurrentEnvVars(projectRoot: string) {
-  const devVarsFile = await Deno.readTextFile(
-    join(projectRoot, envFile),
-  ).catch(() => "");
-  return devVarsFile.split("\n").reduce((acc, line) => {
-    if (!line) {
+export async function getCurrentEnvVars(projectRoot: string): Promise<{
```
igorbrasileiro, line 12: Wdyt about using a lib that only handles parse env file. Example
https://www.npmjs.com/package/dotenv

viktormarinho:
> #### ⚠️ **Areas of Concern**
> 1. **Inconsistent Directory Usage** (`packages/cli/src/hosting/deploy.ts:144`)
>    ```ts
>    const { envVars, envFilepath } = await getCurrentEnvVars(Deno.cwd());
>    ```
>    * **Issue**: Changed from using `cwd` parameter to hardcoded `Deno.cwd()`
>    * **Problem**: This breaks the function's ability to work with different target directories
>    * **Impact**: If a user specifies a different working directory via the `cwd` option, environment variables will be read from the current working directory instead of the specified target
>    * **Recommendation**: Should use `cwd` parameter: `getCurrentEnvVars(cwd)`

This is actually a fix, it wasn't reading the root project directory (where the wrangler file is)

PR #909, opened 2025-07-25
fix: bump @hookform/resolvers to fix schema with formats

igorbrasileiro:
Bump @hookform/resolvers to 5.2.0 which handles ajv-formats allowing schemas with format. Related to https://github.com/react-hook-form/resolvers/pull/797

Example schema: {type: "object", properties: { email: { type: "string", format: "email" } } }

Before: the submit button does nothing, because fails ajvResolver.
After: the submit button make the request.


mcandeia:
gg

PR #933, opened 2025-07-30
Rename menu Integrations to MCPs - welcome MCP CMS!

vibegui:
CleanShot 2025-07-29 at 21 58 49@2x

Next step: "Views" between Agents and Workflows.

### DIFF
--- a/apps/web/src/components/integrations/breadcrumb.tsx
+++ b/apps/web/src/components/integrations/breadcrumb.tsx
@@ -26,7 +26,7 @@ export function IntegrationPageLayout({ tabs }: { tabs: Record<string, Tab> }) {
         hideViewsButton
         breadcrumb={
           <DefaultBreadcrumb
-            items={[{ label: "Integrations", link: "/connections" }]}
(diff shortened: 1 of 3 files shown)


rafavalls (approved):

pedrofrxncx:
"model context protocols"?
i dont think this is good, semantically speaking
could be "MCP Servers" but "Integrations" is still a better name imo

pedrofrxncx:
also there are other places in the codebase mentioning integration(s) / connection(s), they should all be renamed together with this change (but i'm still against it xD)

vibegui:
@lucis @mcandeia @tlgimenes WDYT about @pedrofrxncx remark? MCP Servers? Kind of a mouthful for the Menu

@pedrofrxncx I think "MCPs" are currently being widely used as a plural. Didnt find it strange when I saw it, but maybe im tunneled 

cc @rafaelgcrespo pls advise too!

lucis:
@vibegui @pedrofrxncx I don't think it's a stretch to use MCPs

People are already using it to refer to their servers

Apps is a good option also, but I'm okay with MCP because of the timing

PR #953, opened 2025-08-01
feat: add PDF summarization pre-processing for large files

viktormarinho:
**Add PDF summarization pre-processing for large files**

A user was uploading an 8MB PDF, which consistently failed due to token limits—often even with smaller files. While this issue doesn't occur when uploading the same file directly to ChatGPT, it's likely because OpenAI performs pre-processing to mitigate token overflows.

To address this, we introduced a new flag that enables our chat system to pre-process large PDFs (over 100KB). This pre-processing involves chunking the PDF and summarizing it using a smaller, cheaper model before passing the result to the main LLM agent.

This is a naive first approach to do this pre-processing, and we can do it better later.

### DIFF
--- a/packages/ai/src/agent.ts
+++ b/packages/ai/src/agent.ts
@@ -1064,29 +1064,37 @@ export class AIAgent extends BaseActor<AgentMetadata> implements IIAgent {
         resourceId: options?.resourceId ?? this._thread.resourceId,
       };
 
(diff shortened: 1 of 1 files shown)


viktormarinho:
@claude review

tlgimenes (approved):

PR #992, opened 2025-08-11
Tavano/image as url

guitavano:
before/how chatGPT is/after


viktormarinho (approved):

viktormarinho:
actually thinking better abt this

PR #1050, opened 2025-08-18
feat(runtime): enable drizzle db creation with a fixed db binding

viktormarinho:


packages/runtime/jsr.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@deco/workers-runtime",
-  "version": "0.11.1",
+  "version": "0.12.1",
```
igorbrasileiro: Shouldn't be 0.12.0?

viktormarinho: You're correct, thanks

PR #1080, opened 2025-08-21
refactor: remake apps-auth route to support inline app creation + new ui

viktormarinho:
Note: I still need to work more on the oauth flows, there's some hacky code, i could do some stuff because i can assume all apps are from the registry on the apps-auth screen for now but i will unify this stuff later


viktormarinho:
@claude review

pedrofrxncx:
@claude

PR #1178, opened 2025-09-06
fix error after select first tool

guitavano:
the bug was:

https://github.com/user-attachments/assets/7e1e6951-b5f6-4c65-a2e4-2b4fd42e2693

## Summary by CodeRabbit

- Bug Fixes
  - Updated the toolset selector’s “Select all” checkbox behavior. It now appears checked whenever at least one tool is available, while the toggle still selects or deselects all tools as before. You may see the checkbox checked even if not all tools are enabled; selections only change when you interact with the control.


lucis:
Amazing!!! thank you @guitavano

Issue #1228, opened 2025-09-13
The home page should have "Recent projects" as the first header/section


vibegui:
The home page should have "Recent projects" as the first header/section to allow users to quickly get back to something they were working on instead of having to click twice to find team and project. 

<img width="1685" height="1001" alt="Image" src="https://github.com/user-attachments/assets/39a677b0-3519-4b62-a886-6dbe7bd8b341" />

vibegui:
@claude is the information of user visiting a project being stored? if so, is that feature easy to do? if not, what would it take for us to store visits and have the data for that UX to be possible?

PR #1230, opened 2025-09-14
Add recent projects and fix back button

vibegui:
CleanShot 2025-09-14 at 11 48 34

### Summary
This PR implements the “Recent projects” section at the top of the home page, aggregates activity across all orgs, and aligns the card/UI styles with organization cards. It also fixes avatar/icon sourcing for recent projects and adds last-accessed metadata.

Closes [#1228](https://github.com/deco-cx/chat/issues/1228)

### What’s new
- Recent projects section appears before “My organizations”, not inside the orgs scroller.
- Shows up to 6 most recently accessed projects across any org.
- Each card displays:
  - Path as /org/project
  - Title
  - “Last accessed …” relative time
  - Avatar (org icon fallback if project icon is missing)
  - Subtle border and spacing consistent with org cards

### Backend
- Added PROJECTS_RECENT tool:
  - Reads from user_activity ordered by created_at desc, deduped by team, limit 6.
  - Resolves avatars:
    - Signs relative project icons (team-avatars/...) with the org’s workspace bucket.
    - Includes signed org avatar.
  - Returns last_accessed_at for each item.
- Avoided pulling sandbox/QuickJS tooling into web bundle by excluding sandbox tools from the global exported tools set (prevents Vite dep-scan WASM errors).

### SDK
- CRUD: listRecentProjects()
- Hooks/keys: KEYS.RECENT_PROJECTS(), useRecentProjects() (includes last_accessed_at in project shape)

### Web (apps/web)
- Home/organizations:
  - New RecentProjectsSection rendered above orgs.
  - Cards use same grid, paddings, and subtle border as organization cards.
  - Avatar logic prefers org icon, falls back to project icon; signed absolute URLs.
  - Path text updated to /org and /org/project (removed @ prefix).
  - Headers for Recent projects and My organizations share the same weight/size.
- Minor: kept scroll behavior to only wrap orgs grid.

### Why
- Faster return-to-work for users (requested in #1228).
- Consistent card hierarchy improves scannability.
- Fixes icon inconsistency and relative-path avatar issues.

### Notes on navigation
- Adjusted project landing redirect handling to avoid awkward history behavior when entering a project. Users can navigate back to orgs normally.

### Testing
- Visited multiple projects across orgs, confirmed ordering and dedupe.
- Verified relative project icon paths are signed to absolute URLs.
- Confirmed org icons appear as fallback when project icon missing.
- Checked responsive grid and header sizes.
- Ensured Vite dev server runs without WASM dep-scan errors.

### Risks / Rollout
- Recent projects relies on user_activity; if an environment lacks data, section hides gracefully.
- Tool added to global API surface; scopes validated.

### Checklist
- [x] UI matches design system spacing, borders, and typography
- [x] Uses @deco/ui components
- [x] No linter/type errors
- [x] Docs and code self-explanatory for new tool and hooks

## Summary by CodeRabbit

* **New Features**
  * Recent Projects section on the Organizations page with responsive project cards, avatars, links and “Last accessed” timestamps (includes loading skeleton).

* **Style**
  * Organization cards now have a border.
  * Organization slugs now display as “/slug” instead of “@slug”.

* **Bug Fixes**
  * Navigating from Project Home to Agents now replaces the history entry for better back-button behavior.

* **Improvements**
  * Avatars gracefully fall back on load error.
  * Recent Projects update more reliably when projects are accessed.

### DIFF
--- a/apps/web/src/components/home/organizations.tsx
+++ b/apps/web/src/components/home/organizations.tsx
@@ -136,7 +136,7 @@ function RecentProjectsSection() {
         {recent.map((project) => (
           <Link
             key={`${project.org.slug}/${project.slug}`}
-            to={`/${project.org.slug}/${project.slug}/agents`}
+            to={`/${project.org.slug}/${project.slug}`}
             className="bg-stone-50 hover:bg-stone-100 transition-colors flex flex-col rounded-lg border border-border"
           >
             <div className="p-4 flex flex-col gap-4">
--- a/apps/web/src/components/home/organizations.tsx
+++ b/apps/web/src/components/home/organizations.tsx
@@ -136,7 +136,7 @@ function RecentProjectsSection() {
         {recent.map((project) => (
           <Link
             key={`${project.org.slug}/${project.slug}`}
-            to={`/${project.org.slug}/${project.slug}`}
+            to={`/${project.org.slug}/${project.slug}/agents`}
             className="bg-stone-50 hover:bg-stone-100 transition-colors flex flex-col rounded-lg border border-border"
           >
             <div className="p-4 flex flex-col gap-4">
--- a/apps/web/src/components/home/organizations.tsx
+++ b/apps/web/src/components/home/organizations.tsx
@@ -127,50 +127,78 @@ Organizations.Empty = () => (
 function RecentProjectsSection() {
   const recent = useRecentProjects();
 
-  if (!recent?.length) return null;
-
   return (
     <div className="@container flex flex-col gap-4">
       <h2 className="text-xl font-medium">Recent projects</h2>
-      <div className="grid grid-cols-2 @min-3xl:grid-cols-3 @min-6xl:grid-cols-4 gap-4">
-        {recent.map((project) => (
-          <Link
-            key={`${project.org.slug}/${project.slug}`}
-            to={`/${project.org.slug}/${project.slug}`}
-            className="bg-stone-50 hover:bg-stone-100 transition-colors flex flex-col rounded-lg border border-border"
-          >
-            <div className="p-4 flex flex-col gap-4">
-              <div className="flex justify-between items-start">
-                <Avatar
-                  url={project.avatar_url || project.org.avatar_url || ""}
-                  fallback={project.org.slug}
-                  size="lg"
-                  objectFit="contain"
-                />
-                <Icon
-                  name="chevron_right"
-                  size={20}
-                  className="text-muted-foreground"
-                />
+      {recent?.length ? (
+        <div className="grid grid-cols-2 @min-3xl:grid-cols-3 @min-6xl:grid-cols-4 gap-4">
+          {recent.map((project) => (
+            <Link
+              key={`${project.org.slug}/${project.slug}`}
+              to={`/${project.org.slug}/${project.slug}`}
+              className="bg-stone-50 hover:bg-stone-100 transition-colors flex flex-col rounded-lg border border-border"
+            >
+              <div className="p-4 flex flex-col gap-4">
+                <div className="flex justify-between items-start">
+                  <Avatar
+                    url={project.avatar_url || project.org.avatar_url || ""}
+                    fallback={project.org.slug}
+                    size="lg"
+                    objectFit="contain"
+                  />
+                  <Icon
+                    name="chevron_right"
+                    size={20}
+                    className="text-muted-foreground"
+                  />
+                </div>
+                <div className="flex flex-col gap-[2px]">
+                  <h3 className="text-sm text-muted-foreground truncate">
+                    /{project.org.slug}/{project.slug}
+                  </h3>
+                  <p className="font-medium truncate">{project.title}</p>
+                  {project.last_accessed_at && (
+                    <span className="text-xs text-muted-foreground">
+                      Last accessed {timeAgo(project.last_accessed_at)}
+                    </span>
+                  )}
+                </div>
               </div>
-              <div className="flex flex-col gap-[2px]">
-                <h3 className="text-sm text-muted-foreground truncate">
-                  /{project.org.slug}/{project.slug}
-                </h3>
-                <p className="font-medium truncate">{project.title}</p>
-                {project.last_accessed_at && (
-                  <span className="text-xs text-muted-foreground">
-                    Last accessed {timeAgo(project.last_accessed_at)}
-                  </span>
-                )}
-              </div>
-            </div>
-          </Link>
-        ))}
-      </div>
+            </Link>
+          ))}
+        </div>
+      ) : (
+        <div className="text-sm text-muted-foreground">
+          No recent projects. Create one now.
+        </div>
+      )}
     </div>
   );
 }
+
+RecentProjectsSection.Skeleton = () => (
+  <div className="@container flex flex-col gap-4">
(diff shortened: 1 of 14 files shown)


packages/sdk/src/crud/projects.ts
```diff
@@ -2,3 +2,6 @@ import { MCPClient } from "../fetcher.ts";
 
 export const listProjects = (org: string, init?: RequestInit) =>
   MCPClient.PROJECTS_LIST({ org }, init).then((res) => res.items);
+
+export const listRecentProjects = (init?: RequestInit) =>
+  MCPClient.PROJECTS_RECENT({ limit: 6 }, init).then((res) => res.items);
```
pedrofrxncx: why make a new tool instead of adding filters to the existing one?

vibegui: Short answer: keeping it as a separate tool is the safer, simpler contract.

Why not a filter on PROJECTS_LIST?
- Scope mismatch: PROJECTS_LIST is org-scoped (requires an org slug) and returns all projects for that org. “Recent” is cross-org and returns at most one project per team based on user_activity. Merging both behaviors into one tool would either:
  - Overload semantics (org-specific vs cross-org), or
  - Force an “org = *” convention that breaks the current API shape and cache keys.
- Caching/keys: the SDK caches by `KEYS.PROJECTS(org)`. A cross-org “recent” variant would not fit that cache key cleanly and risks wrong cache hits/invalidations. Using `KEYS.RECENT_PROJECTS()` keeps caching straightforward and isolated.
- Auth and joins: PROJECTS_RECENT uses user_activity and a members join to dedupe and authorize across orgs, which is a different access pattern than PROJECTS_LIST’s single-org membership check.

If we really want a single endpoint later:
- We could introduce a new, clearly separate tool like PROJECTS_QUERY with a typed filter object (e.g., mode: "by_org" | "recent", org?: string, limit?: number). That preserves clarity while reducing surface area, but it’s more work now and doesn’t improve the current UX/code much.

packages/sdk/src/mcp/teams/api.ts
```diff
@@ -1235,3 +1235,161 @@ export const listProjects = createTool({
...
+      }),
+    ),
+  }),
+  handler: async (props, c) => {
+    assertPrincipalIsUser(c);
+    c.resourceAccess.grant();
+
+    const user = c.user;
+    const { limit } = props;
+    const safeLimit = Math.min(24, Math.max(1, Number(limit ?? 12)));
```
viktormarinho: zod is already validating this on the schema

packages/sdk/src/mcp/teams/api.ts
```diff
@@ -1235,3 +1235,161 @@ export const listProjects = createTool({
...
+
+    // Get latest user activity rows (most recent first) and dedupe by team id
+    const { data: activityData, error: activityError } = await c.db
+      .from("user_activity")
+      .select("value, created_at")
+      .eq("user_id", user.id)
+      .eq("resource", "team")
+      .order("created_at", { ascending: false })
+      .limit(Math.min(200, Math.max(50, safeLimit * 4)));
```
viktormarinho: weird that we do this here then below deduplicate in memory, this should be done 100% on the query putting the exact limit here + using DISTINCT

packages/sdk/src/mcp/teams/api.ts
```diff
@@ -1235,3 +1235,161 @@ export const listProjects = createTool({
...
+        )
+      `,
+      )
+      .in("org_id", orderedTeamIds)
+      .eq("teams.members.user_id", user.id)
+      .is("teams.members.deleted_at", null);
+
+    if (projectsError) throw projectsError;
+
+    // Group projects by team id and pick one (prefer slug 'default')
```
viktormarinho: Also, the resource that we're checking for activity should be a *project*, not a team then using some logic to get a project from that team. 

This will simplify the logic and also fix potential bugs since today you can only have "personal" or "default" as project slugs but it will change soon once we're able to create more projects and this will stop making sense as well

packages/sdk/src/hooks/teams.ts
```diff
@@ -84,6 +75,21 @@ export const useProjects = (options: {
...
+  const query = useSuspenseQuery<Project[]>({
+    queryKey: KEYS.RECENT_PROJECTS(),
+    queryFn: () => listRecentProjects(),
+    // Reduce auth/mcp pressure: rely on cache, no automatic refetches
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    networkMode: "online",
+    staleTime: 60_000,
+    gcTime: 300_000,
+    retry: 0,
```
viktormarinho:
```suggestion
```

Default stale/refetching behaviors seems good UX for this query. i see no need to make those overrides actually, besides only a disable for the refetch on mount in case there is some weird rerendering going on.

packages/sdk/src/hooks/teams.ts
```diff
@@ -53,17 +54,7 @@ export const useOrganizations = (options: { searchQuery?: string } = {}) => {
...
-  title: string;
-  slug: string;
-  avatar_url: string | null;
-  org: {
-    id: number;
-    slug: string;
-    avatar_url?: string;
-  };
-}
+// Project interface moved to ../models/project.ts to avoid drift
```
viktormarinho:
```suggestion
```

apps/web/src/components/home/organizations.tsx
```diff
@@ -131,6 +180,7 @@ function MyOrganizations() {
     <div className="flex w-full h-full items-start bg-background">
       <div className="p-8 flex flex-col gap-4 w-full">
         <DecoDayBanner />
+        <RecentProjectsSection />
```
viktormarinho: make a skeleton (you can reuse the same as org/project listing prob)

and wrap this with suspense/error boundary. since you're using useSuspenseQuery putting this here without wrapping kills our current loading behavior with skeletons for this whole component tree.

apps/web/src/components/home/organizations.tsx
```diff
@@ -123,6 +124,54 @@ Organizations.Empty = () => (
...
+  if (!recent?.length) return null;
+
+  return (
+    <div className="@container flex flex-col gap-4">
+      <h2 className="text-xl font-medium">Recent projects</h2>
+      <div className="grid grid-cols-2 @min-3xl:grid-cols-3 @min-6xl:grid-cols-4 gap-4">
+        {recent.map((project) => (
+          <Link
+            key={`${project.org.slug}/${project.slug}`}
+            to={`/${project.org.slug}/${project.slug}/agents`}
```
viktormarinho:
```suggestion
            to={`/${project.org.slug}/${project.slug}`}
```

I would keep the redirect to the homepage, since it already redirects to /agents today, and we can change this later in only one place. The homepage can be another things, as well.

viktormarinho (changes requested):

packages/sdk/src/mcp/teams/api.ts
viktormarinho: maybe you'll have to create a procedure on the supabase dashboard to do this, 😅

we should move to query with drizzle orm soon

apps/web/src/components/home/organizations.tsx
vibegui: You know GPT suggested this and I thought he was hallucinating and being overkill lol now I know

PR #1345, opened 2025-10-02
chore: add deco_tool_run_tool

tlgimenes:


lucis (commented):
@tlgimenes did you check for authorization of this API?

e.g: users bypassing role-based security by calling tools via dynamically generated code

PR #1464, opened 2025-10-14
avoid show legacy option to users that dont need

guitavano:
Legacy option should be showed only to users that have at least one document in the old model.

Before:
Screenshot 2025-10-14 at 00 37 57

After:
Screenshot 2025-10-14 at 00 38 47


guitavano:
Change to use user preferences

PR #1539, opened 2025-10-16
Fix(cli): Duplicated "Env" type definition

vitoUwu:
This fixes a bug where some app tool could have a property named as "env" and this would create a type/interface named as Env which conflicts with the default Env interface generated by Deco. This was happening because the gen command have a map for every type/interface created with a count to avoid conflicts, to avoid conflicts with "Env" interface we populate it with `types.set("Env", 1);`, to check if we already have a count for a specific type name we format the name and check if the values exists inside the map

```ts
const key = slugify(typeName);
const count = types.get(key) ?? 0;
```

The slugify function receive the name and transform it to lowercase
```ts
function slugify(name: string) {
  return name.replace(/[^a-zA-Z0-9]/g, "").toLowerCase();
}
```
When checking for "env' type, the count don't exists yet because we define the initial count to "Env", lowercasing the initial count fixes the problem.

## Screenshots/Demonstration
Before:
image
After:
image

## Review Checklist
- [x] No breaking changes 

## Summary by CodeRabbit

* **Chores**
  * CLI package version updated to 0.22.5.

### DIFF
--- a/packages/cli/src/commands/gen/gen.ts
+++ b/packages/cli/src/commands/gen/gen.ts
@@ -200,6 +200,7 @@ export const genEnv = async ({
   try {
     const types = new Map<string, number>();
     types.set("env", 1); // set the default env type
+    types.set("Env", 1);
     let tsTypes = "";
     const mapBindingTools: Record<string, string[]> = {};
     const props = await Promise.all(


packages/cli/src/commands/gen/gen.ts
```diff
@@ -199,7 +199,7 @@ export const genEnv = async ({
 
   try {
     const types = new Map<string, number>();
-    types.set("Env", 1); // set the default env type
+    types.set("env", 1); // set the default env type
```
mcandeia, line 202: you need to duplicate this line
```suggestion
    types.set("env", 1); // set the default env type
    types.set("Env", 1);
```

PR #1702, opened 2025-10-30
feat: step views

pedrofrxncx:
step views generation with decopilot

## Summary by CodeRabbit

* **New Features**
  * "Create view" button added to workflow step outputs for quick view creation.
  * Chat/thread management improved for more reliable thread selection and remount behavior.

* **Improvements**
  * Step output layout refreshed with horizontal, scrollable views and improved controls; step name now shown in outputs.
  * Save/workflow flows simplified with unified save behavior, clearer button states, consistent success toasts, and preserved workflow URIs.
  * Input blur now saves only when content changed.

* **Bug Fixes**
  * Context item identifiers guaranteed when missing to avoid collisions.

### DIFF
--- a/apps/web/src/stores/workflows/hooks.ts
+++ b/apps/web/src/stores/workflows/hooks.ts
@@ -118,10 +118,10 @@ export function useWorkflowStepData(stepName: string) {
     () => (state: Store) => ({
       output:
         state.stepOutputs[stepName] ?? state.stepOutputs[stepNameWithoutId],
-      views: state.stepOutputs[stepNameWithoutId]
-        ? (state.stepOutputs[stepNameWithoutId] as { views?: string[] }).views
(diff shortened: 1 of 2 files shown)


apps/web/src/components/workflow-builder/steps/output.tsx
```diff
@@ -49,43 +59,140 @@ function deepParse(value: unknown, depth = 0): unknown {
...
+  } as RuleContextItem;
+};
+
+function CreateViewButton({ stepName }: { stepName: string }) {
+  const step = useWorkflowStepData(stepName);
+  const { setOpen } = useDecopilotOpen();
+  const { setThreadState } = useDecopilotThread();
+  const { contextItems, setContextItems } = useThreadContext();
+
+  const defaultContextItems = useMemo(
```
tlgimenes: why not already add these rules and tools on the api directly instead of on the UI?

PR #1726, opened 2025-10-30
chore: deny_list for purplevara

tlgimenes:
## Summary by CodeRabbit

* **Chores**
  * Restricted the list of available integrations shown in the app by filtering out specific integrations; these denied options are no longer visible when browsing or selecting integrations.

### DIFF
--- a/apps/api/src/decopilot-stream.ts
+++ b/apps/api/src/decopilot-stream.ts
@@ -198,7 +198,6 @@ type ToolLike = {
 
 const INTEGRATIONS_DENY_LIST = new Set([
   "i:contracts-management",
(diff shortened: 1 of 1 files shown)


apps/api/src/decopilot-stream.ts
```diff
@@ -196,6 +196,23 @@ type ToolLike = {
   outputSchema?: Record<string, unknown>;
 };
 
+const INTEGRATIONS_DENY_LIST = new Set([
+  "i:contracts-management",
+  "i:thread-management",
```
vibegui: i would leave threads

PR #1790, opened 2025-11-06
feat: add supportsToolName option to CreateStubAPIOptions and update MCPClientProxy logic

kevinicolas22:
## Summary by CodeRabbit

* **New Features**
  * Added an optional configuration to control whether tool names are included in API request paths.
  * Added automatic endpoint detection to conditionally include tool names for compatible services; explicit configuration takes precedence over detection.

### DIFF
--- a/packages/runtime/src/proxy.ts
+++ b/packages/runtime/src/proxy.ts
@@ -4,6 +4,7 @@ import { convertJsonSchemaToZod } from "zod-from-json-schema";
 import { MCPConnection } from "./connection.ts";
 import { createServerClient } from "./mcp-client.ts";
 import type { CreateStubAPIOptions } from "./mcp.ts";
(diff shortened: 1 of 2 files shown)


packages/runtime/src/proxy.ts
```diff
@@ -32,6 +32,28 @@ const toolsMap = new Map<
...
+  try {
+    const urlObj = new URL(url);
+    const hostname = urlObj.hostname.toLowerCase();
+
+    // Our main APIs that support /tool/${toolName} routing
+    return (
+      hostname === "api.decocms.com" ||
+      hostname === "api.deco.chat" ||
+      hostname === "localhost" ||
+      hostname.startsWith("localhost") // covers localhost:3001, etc.
```
mcandeia: nit: startsWith("localhost") already covers the === "localhost" case

PR #1843, opened 2025-11-12
Feat/improve chat

viniciusventura29:
An UX/UI improviment in deco chat :) 

## Screenshots/Demonstration
To see what I am proposing, watch this: https://www.loom.com/share/d7ca3e940f1447bd9b7e732a9d224722

## Summary by CodeRabbit

* **New Features**
  * Chat now groups messages into question–answer pairs with a compact pair view and clickable headers to focus a pair.
  * Rich message rendering: images, inline previews, attachments with lazy text preview, and copy-to-clipboard.
  * Expand/collapse for long messages and hover actions for copy.

* **Bug Fixes**
  * Improved scroll behavior to focus the latest pair on submit.
  * Updated loading/streaming indicator behavior for last messages.

### DIFF
--- a/apps/web/src/components/chat/assistant-chat-message.tsx
+++ b/apps/web/src/components/chat/assistant-chat-message.tsx
@@ -50,10 +50,10 @@ export function AssistantChatMessage({
     !message?.parts &&
     (status === "streaming" || status === "submitted");
 
-    const minHeightClass = useMemo(() => {
-      if (!isLastMessage) return "";
-      return hasTabs ? "min-h-[63vh]" : "min-h-[69vh]";
-    }, [isLastMessage, hasTabs]);
+  const minHeightClass = useMemo(() => {
+    if (!isLastMessage) return "";
+    return hasTabs ? "min-h-[63vh]" : "min-h-[69vh]";
+  }, [isLastMessage, hasTabs]);
 
   return (
     <div
--- a/apps/web/src/components/chat/chat-pair-messages.tsx
+++ b/apps/web/src/components/chat/chat-pair-messages.tsx
@@ -51,7 +51,11 @@ export const ChatAskAnswerPair = forwardRef<
       </div>
 
       <div className="message-block px-4 py-4">
-        <AssistantChatMessage hasTabs={hasTabs} message={assistant} isLastMessage={isLastPair} />
+        <AssistantChatMessage
+          hasTabs={hasTabs}
+          message={assistant}
+          isLastMessage={isLastPair}
+        />
       </div>
     </div>
   );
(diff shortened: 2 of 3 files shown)


apps/web/src/components/chat/assistent-chat-message.tsx
```diff
@@ -0,0 +1,375 @@
...
+      .then((t) => {
+        if (!cancelled) setText(t);
+      })
+      .catch(() => {
+        if (!cancelled) setText(null);
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [attachment.url, expanded, text]);
```
lucis: Potential for frontend bug

apps/web/src/components/chat/assistent-chat-message.tsx
```diff
@@ -0,0 +1,375 @@
...
+}
+
+function AttachmentCard({ attachment }: { attachment: MessageAttachment }) {
+  const contentType =
+    attachment.contentType ||
+    (attachment.url.startsWith("data:")
+      ? attachment.url.split(":")[1]?.split(";")[0]
+      : undefined) ||
+    "application/octet-stream";
+  const isImage = contentType.startsWith("image/");
```
lucis: Do we really need all of this logic?

I'd expect for us to already have that somewhere

apps/web/src/components/chat/chat-messages.tsx
```diff
@@ -1,68 +1,55 @@
...
-  );
-}
-
 export function ChatMessages({ className }: ChatMessagesProps = {}) {
   const sentinelRef = useRef<HTMLDivElement | null>(null);
   const {
     chat: { messages, status },
   } = useAgenticChat();
 
+  const formattedAskAnswerMessages = useMemo(
```
lucis: What's the meaning of AskAnswerMessages?

What's "AskAnswer"?

Are there message which are not "AskAnswer"? Which ones?

viniciusventura29: I misspoke… I used the name ‘MessagesInPairs’, which is exactly what I meant... I think it fits better

apps/web/src/components/chat/assistent-chat-message.tsx
viniciusventura29: We needed this, but the logic in this file was very bad and confusing, so I rewrote it to be more readable and moved everything to another file

From my understanding, this logic is a robust way to handle attachments when `attachment.contentType` is not filled so beacuse of this we need but if you think different, please let me know

apps/web/src/components/chat/assistent-chat-message.tsx
viniciusventura29: Yes, this smell a little. Have an idea how to improve it? 

This is our current implementation, I just moved it from one file to another

PR #2002, opened 2025-12-13
First event bus version

mcandeia:
## Review Checklist
- [ ] No breaking changes 

---
## Summary by cubic
Introduces the first MCP Mesh event bus using CloudEvents with at-least-once delivery, batching, retries, and scheduled and recurring (cron) delivery. Uses PostgreSQL LISTEN/NOTIFY or SQLite polling and auto-starts a background worker on app boot.

- **New Features**
  - Event bus with publish/subscribe APIs, per-subscription delivery tracking, batching via ON_EVENTS, retry/backoff, deliverAt scheduling, cron for recurring events, and EVENT_ACK for explicit acknowledgments after retryAfter.
  - PostgreSQL LISTEN/NOTIFY integration with polling fallback; SQLite uses polling.
  - New MCP tools: EVENT_PUBLISH, EVENT_SUBSCRIBE, EVENT_UNSUBSCRIBE, EVENT_CANCEL, EVENT_ACK, EVENT_SUBSCRIPTION_LIST.
  - New well-known bindings: EventBusBinding and EventSubscriberBinding (CloudEvents schemas).
  - App wiring: event bus created in createApp, worker auto-started; MeshContext exposes eventBus and caller connectionId (propagated via x-caller-id).

- **Migration**
  - Run migration 008-event-bus to create events, event_subscriptions, and event_deliveries tables.
  - Connections that receive events must implement ON_EVENTS per EventSubscriberBinding.
  - Use connection-scoped tokens when publishing/subscribing so the caller connectionId is available.

<sup>Written for commit d91521e4ad2bf1ab8fac592ec4d4f0551f299c9e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/mesh/src/event-bus/index.ts
+++ b/apps/mesh/src/event-bus/index.ts
@@ -6,12 +6,14 @@
  * Architecture:
  * - EventBus: Single class handling publish/subscribe and worker management
  * - EventBusStorage: Database operations (unified for SQLite/PostgreSQL via Kysely)
- * - EventBusWorker: Polling and delivery logic
- * - NotifyStrategy: Optional immediate wake-up (PostgreSQL uses LISTEN/NOTIFY)
+ * - EventBusWorker: Event processing and delivery logic (no internal polling)
+ * - NotifyStrategy: Triggers worker processing
+ *   - SQLite: Timer-based polling
+ *   - PostgreSQL: Event-based via LISTEN/NOTIFY
  *
  * Usage:
  * ```ts
- * const eventBus = createEventBus(database, notifySubscriber, config);
+ * const eventBus = createEventBus(database, config);
  * await eventBus.start();
  * ```
  */
@@ -19,7 +21,12 @@
 import type { MeshDatabase } from "../database";
 import { createEventBusStorage } from "../storage/event-bus";
 import { EventBus as EventBusImpl } from "./event-bus";
-import type { EventBus, EventBusConfig } from "./interface";
+import {
+  DEFAULT_EVENT_BUS_CONFIG,
+  type EventBus,
+  type EventBusConfig,
+} from "./interface";
+import { PollingStrategy } from "./polling";
 import { PostgresNotifyStrategy } from "./postgres-notify";
 
 // Re-export types and interfaces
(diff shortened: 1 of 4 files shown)


apps/mesh/src/api/index.ts
```diff
@@ -72,6 +79,34 @@ export interface CreateAppOptions {
...
+        }),
+      });
+      const ctx = await ContextFactory.create(systemRequest);
+      return {
+        ...ctx,
+        auth: { ...ctx.auth, user: { id: "notify-worker" } },
+      };
+    });
+
+    eventBus = createEventBus(db, notifySubscriber, undefined, databaseUrl);
```
viktormarinho: we should be able to create a context without a request

apps/mesh/src/api/index.ts
```diff
@@ -321,5 +362,19 @@ export function createApp(options: CreateAppOptions = {}) {
...
+  }
+  return defaultApp;
+}
+
+// Re-export for backwards compatibility (but now lazy)
+export default new Proxy({} as ReturnType<typeof createApp>, {
+  get(_, prop) {
+    return (getDefaultApp() as never)[prop];
+  },
+});
```
viktormarinho: we should find a way to structure the code without needing this just for tests to work

apps/mesh/src/event-bus/index.ts
```diff
@@ -0,0 +1,61 @@
...
+function detectDatabaseType(databaseUrl?: string): DatabaseType {
+  if (!databaseUrl) return "sqlite";
+
+  const url = databaseUrl.toLowerCase();
+  if (url.startsWith("postgres://") || url.startsWith("postgresql://")) {
+    return "postgres";
+  }
+  return "sqlite";
+}
+
```
viktormarinho: i think we already have this code on databases module

apps/mesh/src/event-bus/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  databaseUrl?: string,
+): EventBus {
+  const dbType = detectDatabaseType(databaseUrl);
+
+  if (dbType === "postgres") {
+    return new PostgresEventBus(db, notifySubscriber, config, databaseUrl);
+  }
+
+  return new SqliteEventBus(db, notifySubscriber, config);
+}
```
viktormarinho, line 77: weird that we receive both laready instantiated kysely db and database url then we detect which one we are using to then instantiate different buses based on db. we can make this better

PR #2530, opened 2026-03-01
feat(ui): Model selector overhaul

rafavalls:
Two files changed: context.tsx and select-model.tsx.

context.tsx — ModelAutoSelector now prefers Claude Opus 4.6 → Sonnet 4.6 as default instead of taking the first model in the list.

select-model.tsx — Full redesign of the model picker:

Popover → Dialog — centered modal instead of anchored popover, no more viewport overflow issues
Tier system — models grouped into Smarter / Faster / Cheaper via prefix matching on model IDs; unrecognized models fall to “Other”
Shortlist — users pick which models appear in the list; persisted per-connection in localStorage via mesh:model-shortlist:<connectionId>; browse mode shows shortlist, manage mode shows all with checkboxes
Details panel — model logo + name header, capability pills, context window dots (red→green by absolute thresholds), output limit with word approximation, pricing $$$$ indicators (green→red)
Removed — percentile badges (Top 10% etc.), XL/L/M/S size labels, per-row cost display; all were relative to the loaded list.

### DIFF
--- a/apps/mesh/src/web/components/chat/select-model.tsx
+++ b/apps/mesh/src/web/components/chat/select-model.tsx
@@ -117,10 +117,13 @@ function classifyModel(modelId: string): TierId | null {
   if (modelId.endsWith(FREE_SUFFIX)) return "cheaper";
   for (const { tier, prefix } of SORTED_TIER_RULES) {
     if (modelId.startsWith(prefix)) {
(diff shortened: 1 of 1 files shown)


rafavalls:
@cubic-dev-ai rewrite the pr description pls?

PR #2528, opened 2026-02-28
feat: local-first DX with zero-ceremony setup and model selector overhaul

vibegui:
## Summary

- **Local-first setup**: `bunx @decocms/mesh` creates a `~/deco` data directory, auto-provisions a local admin user (using OS username), seeds the database, and starts the server — zero config required
- **Auto-auth connections**: Installing a provider (e.g. OpenRouter) triggers automatic OAuth authentication and redirects back to home instead of the connection detail page
- **Model selector overhaul**: Complete redesign of the model picker as a centered Dialog modal with tier-based categorization, shortlist management, and rich comparison tools

## Model Selector Details

- **Tier categories**: Models classified into Smarter / Faster / Cheaper using selective pattern matching on model IDs (flagship models only, variants go to "Other")
- **Shortlist management**: "Manage models" view with per-tier checkboxes, persisted per-connection in localStorage. Default shortlist covers Anthropic, OpenAI, Google, xAI, Cohere, Mistral, Qwen, MiniMax
- **Rich details panel**: Cost color coding (green→red) with percentile tags (Bottom 10%, Budget, Premium, Top 5%), token size tags (Small/Medium/Large/XL), model ID display, vendor/name split header
- **UX improvements**: Centered Dialog instead of Popover (no viewport overflow), always-visible model name on desktop, default to selected model details on open, alphabetical sorting within tiers

## Local Mode Details

- CLI prompts for data directory on first run, persists secrets in `~/deco/secrets.json`
- Auto-builds frontend if `dist/client/` is missing
- Dev script (`scripts/dev.ts`) mirrors CLI environment with Vite HMR
- Local admin uses DB-backed role with auto-redirect to first org

## Test plan

- [ ] `bunx @decocms/mesh` from scratch (no `~/deco`) → prompts for directory, starts server, auto-login works
- [ ] Click "Install OpenRouter" → auto-authenticates, returns to home
- [ ] Open model selector → Smarter/Faster/Cheaper tiers, default model pre-selected in details panel
- [ ] Click "Manage models" → grouped checkboxes, hover shows details with price/percentile tags
- [ ] Toggle models → persists across page refresh
- [ ] `bun run fmt && bun run check && bun run lint` all pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2541, opened 2026-03-02
fix(storage): replace .returningAll() with insert+select for SQLite compat

viktormarinho:
## Summary

- `kysely-bun-worker` (SQLite) does not support the `RETURNING` clause
- Replace all `INSERT/UPDATE … RETURNING *` with a separate follow-up `SELECT` so storage operations work on both SQLite and PostgreSQL

**Files changed:**
- `apps/mesh/src/storage/threads.ts`
- `packages/mesh-plugin-workflows/server/storage/workflow-collection.ts`
- `packages/mesh-plugin-workflows/server/storage/workflow-execution.ts`

## Test plan

- [ ] Run existing workflow engine tests: `bun test packages/mesh-plugin-workflows`
- [ ] Verify thread creation works end-to-end against SQLite (`bun run dev`)
- [ ] Verify workflow create/update/delete/cancel/resume still work against PostgreSQL

## Known issues flagged in review

The SELECT+UPDATE pattern introduces TOCTOU windows that the original atomic `RETURNING` prevented. Specifically: `cancelExecution`, `claimExecution`, `createStepResult`, and `updateExecution` can return incorrect results under concurrent load. These regressions should be addressed with `numUpdatedRows` checks in a follow-up — tracked separately to keep this PR scoped to the compat fix.

> Split from #2528

---
## Summary by cubic
Restored atomic INSERT/UPDATE … RETURNING queries across storage to fix concurrency issues and confirm compatibility with SQLite 3.35+ and PostgreSQL.

- **Bug Fixes**
  - Reverted insert/update + select back to RETURNING in threads, workflow collection, and workflow execution.
  - Fixed concurrent correctness: step-result claiming and conditional updates now respect affected rows atomically.
  - Confirmed BunWorkerDialect supports RETURNING; no SQLite fallback needed.

<sup>Written for commit d8856d49aad11c93049f4bd34764bcf34ae73f52. Summary will update on new commits.</sup>

### DIFF
--- a/apps/mesh/src/storage/threads.ts
+++ b/apps/mesh/src/storage/threads.ts
@@ -52,13 +52,10 @@ export class SqlThreadStorage implements ThreadStoragePort {
       updated_by: data.updated_by ?? null,
     };
 
-    // Insert the thread, then select it back (compatible with both PostgreSQL and SQLite)
-    await this.db.insertInto("threads").values(row).execute();
-
     const result = await this.db
-      .selectFrom("threads")
-      .selectAll()
-      .where("id", "=", id)
+      .insertInto("threads")
+      .values(row)
+      .returningAll()
       .executeTakeFirstOrThrow();
 
     return this.threadFromDbRow(result);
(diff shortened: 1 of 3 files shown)


viktormarinho:
does not make sense

PR #2549, opened 2026-03-02
Fix login UX: auto-redirect to last org

rafavalls:

Fixes two frustrating login UX issues:
1. **Auto-redirect to last org** — After login, users are immediately redirected to their last active organization instead of seeing a manual org selection screen
2. **Fresh chat on new session** — Chat history no longer persists across browser sessions, so users see a clean empty chat after logging in rather than resuming old conversations

## How to Test

1. **Test org auto-redirect:**
   - Log out completely
   - Log back in → should be automatically directed to the last org you were in (or your only org if you have just one)
   - Navigation to `/` still shows the org list if you explicitly want to switch

2. **Test fresh chat:**
   - Open the chat panel and send a message to create a thread
   - Close the browser tab/window completely (clear the session)
   - Log back in → chat should be empty/fresh, not showing the old conversation
   - Within the same browser session, navigate away and back → the thread is restored

## Implementation Details

- **Fix 1:** Added `authClient.useActiveOrganization()` check in `OrganizationsHome`. Renders `<Navigate>` redirect when data is loaded. Falls back to auto-redirect if user only has one org.
- **Fix 2:** Created new `useSessionStorage` hook (mirrors `useLocalStorage` but uses `sessionStorage`). Switched active thread ID storage to this hook, which clears when the browser session ends.

## Review Checklist
- [x] No breaking changes
- [x] Type checking passes (`bun run check`)

---
## Summary by cubic
Auto-redirects to your last (or only) org after login with safeguards against loops and transient failures. Chats keep the current thread across refreshes; the app version is shown again in the user menu and settings.

- **Bug Fixes**
  - Home route `beforeLoad`: validates a cached `lastOrgSlug`, clears stale values, skips validation if org list fetch fails, and falls back to a single-org redirect; only writes the slug after a successful `setActive`. E2E now reads the slug from `/$org/org-admin`.
  - After sign-in or sign-up, the auth form always navigates to `/` to trigger the redirect flow; handles a null `redirectUrl` to avoid navigating to the string "null".
  - Persist the active chat thread in `localStorage`; removed the unused session-storage hook.

<sup>Written for commit 80fbc421a5ba9f1c19d51e9229e3a61770d37661. Summary will update on new commits.</sup>

### DIFF
--- a/apps/mesh/src/web/components/unified-auth-form.tsx
+++ b/apps/mesh/src/web/components/unified-auth-form.tsx
@@ -16,7 +16,7 @@ interface UnifiedAuthFormProps {
 
 type FormView = "signIn" | "signUp" | "forgotPassword";
 
-export function UnifiedAuthForm({ redirectUrl = "/" }: UnifiedAuthFormProps) {
+export function UnifiedAuthForm({ redirectUrl }: UnifiedAuthFormProps) {
   const { emailAndPassword, resetPassword } = useAuthConfig();
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
@@ -65,7 +65,7 @@ export function UnifiedAuthForm({ redirectUrl = "/" }: UnifiedAuthFormProps) {
     },
     onSuccess: () => {
       globalThis.localStorage?.setItem("hasLoggedIn", "true");
-      window.location.href = redirectUrl;
+      window.location.href = redirectUrl ?? "/";
     },
   });
 


apps/mesh/src/web/components/unified-auth-form.tsx
```diff
@@ -67,6 +67,10 @@ export function UnifiedAuthForm({ redirectUrl }: UnifiedAuthFormProps) {
       globalThis.localStorage?.setItem("hasLoggedIn", "true");
       if (redirectUrl) {
         window.location.href = redirectUrl;
```
viktormarinho: just make the default redirect url be /

PR #2638, opened 2026-03-10
fix(release): use correct GHCR image path for mesh docker image

tlgimenes:
## Summary
- `IMAGE_NAME` was set to `${{ github.repository }}/mesh` which resolves to `decocms/studio/mesh` (since this is the `studio` monorepo)
- The helm chart (`argo-deploy-mesh`) expects the image at `ghcr.io/decocms/mesh/mesh`
- This mismatch caused pods to fail with `ImagePullBackOff` since the tag was never pushed to the expected registry path

## Fix
Hardcode `IMAGE_NAME: decocms/mesh/mesh` so the Docker image is pushed to the correct GHCR path.

## Test plan
- [ ] Merge and verify the release workflow pushes to `ghcr.io/decocms/mesh/mesh`
- [ ] Verify pods come up after next release

---
## Summary by cubic
Reverts the previous change to the Mesh release workflow. Removes the hardcoded `IMAGE_NAME: decocms/mesh/mesh` and restores `${{ github.repository }}/mesh` so images continue publishing to `ghcr.io/decocms/studio/mesh`.

<sup>Written for commit 531e20fad6db6da46f0337454a50939c0240a092. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: decocms/mesh/mesh
+  IMAGE_NAME: ${{ github.repository }}/mesh
 
 jobs:
   release:


tlgimenes:
Closing — the fix should be in the infra repos, not here.

PR #2912, opened 2026-03-27
feat(ui): align sidebar icons and improve panel interactions

rafavalls:

Aligns sidebar icons and improves panel interactions in the shell layout:

- **Sidebar icon alignment**: Adjusts padding and spacing so sidebar toggle/action icons are visually aligned
- **New task shortcut**: Adds a "New task" button (with tooltip showing ⇧⌘S) visible when the tasks panel is collapsed, plus a global keyboard shortcut
- **Tasks panel improvements**: Removes fixed `maxSize` constraint so the tasks panel can be resized wider; renames to `TasksResizablePanel` since width is no longer persisted
- **Debounced inline editing**: Refactors SpaceIdentityHeader to use `useState` + `useEffect` debounce pattern (controlled inputs) instead of `useRef`-based timers
- **Task list styling**: Uses `cn()` for conditional className composition in tasks-panel
- **Tasks filtering**: Passes `agentVirtualMcpId` on agent routes so the tasks panel filters to the current agent's tasks

## Screenshots/Demonstration

N/A

## How to Test

1. Open the app and navigate to an agent route
2. Verify sidebar icons (tasks toggle, chat toggle) are vertically aligned
3. Collapse the tasks panel — a "New task" button with pencil icon should appear with tooltip showing the shortcut
4. Press ⇧⌘S (or ⇧Ctrl+S on non-Mac) — should trigger new task creation
5. Open tasks panel and edit a space name/description — changes should debounce and save after 1s of inactivity
6. Resize the tasks panel wider than 22% — should now be allowed

## Migration Notes

N/A

## Review Checklist
- [x] No breaking changes

### DIFF
--- a/apps/mesh/src/web/layouts/shell-layout.tsx
+++ b/apps/mesh/src/web/layouts/shell-layout.tsx
@@ -520,7 +520,7 @@ function ShellLayoutInner({
           className="flex flex-col"
           style={{ background: "transparent", containerType: "inline-size" }}
         >
-          <div className="shrink-0 flex items-center justify-between pl-5 pr-2 h-10">
+          <div className="shrink-0 flex items-center justify-between pl-1 pr-2 h-10">
             <div className="flex items-center gap-0.5 min-w-0">
               {showThreePanels && (
                 <>


apps/mesh/src/web/components/chat/side-panel-tasks.tsx
```diff
@@ -156,6 +176,15 @@ function SpaceIdentityHeader({ project }: { project: VirtualMCPEntity }) {
     }
   };
 
+  // oxlint-disable-next-line ban-use-effect/ban-use-effect — clears pending debounce timers on unmount; no React 19 alternative for cleanup
+  useEffect(() => {
```
tlgimenes: I'd rather useEffect for the debounce uppdate instead of relying on the timer. Something like

const [title, setTitlte] = useState()

handleTitleChange = (e) => setTitle(e)

useEffect(() => {
   const timer = setTimeout(() => actions.update.mutate, 1000)
  
    return () => {
       canlceTimeot(timer)
  }
}, [title])

tlgimenes: Refactored to use `useState` + `useEffect` debounce pattern as suggested. Each field now has its own effect that sets a timeout on value change and cleans up on re-render/unmount.

PR #2943, opened 2026-03-30
fix(automations): handle NATS 503 no-responders during reconnection

tlgimenes:

Fixes `NatsError: no responders (503)` errors flooding logs when NATS reconnects during `AutomationJobStream` consumer fetch.

**Root cause:** When NATS reconnects, `init()` is called again but the old consumer loop keeps running with a stale reference, producing 503 errors. Setting `this.running = false` at the start of `init()` stops the stale loop so `startConsumer()` can start fresh.

Additionally, 503 errors are downgraded from `error` to `warn` since they are expected transient states during reconnection.

## How to Test

1. Deploy to an environment with NATS
2. Trigger a NATS reconnection (e.g., restart NATS server)
3. Observe logs — should see `warn` level "No responders (NATS reconnecting)" instead of `error` level stack traces
4. Consumer should resume processing normally after reconnection

## Review Checklist
- [x] No breaking changes

---
## Summary by cubic
Switches automation jobs to a singleton job stream using NATS `.consume()` and improves reconnection handling to prevent duplicate consumers and 503 "no responders". Also re-fires `onReady` callbacks on NATS reconnect and resets pod heartbeat so JetStream resources are re-created cleanly.

- **Refactors**
  - Replace manual fetch loop with `.consume()`; `init()` stops any prior subscription for clean reconnects.
  - Expose `automationJobStream` module with `init/publish/startConsumer/stop`; app/tests call `automationJobStream.init(natsOpts)` and inject `automationJobStream.publish` into `AutomationCronWorker`.

- **Bug Fixes**
  - NATS provider listens for `Reconnect` and re-fires `onReady` callbacks; invalidates cached JetStream client.
  - `NatsPodHeartbeat` clears stale watchers/timers and re-inits on reconnect.

<sup>Written for commit ad6aa6c557a3abfb2ca66492c85ef78dcaf885a5. Summary will update on new commits.</sup>

### DIFF
--- a/apps/mesh/src/nats/connection.ts
+++ b/apps/mesh/src/nats/connection.ts
@@ -11,7 +11,12 @@
  * Consumers should use onReady() to defer work until the connection is available.
  */
 
-import { connect, type JetStreamClient, type NatsConnection } from "nats";
+import {
+  Events,
+  connect,
+  type JetStreamClient,
+  type NatsConnection,
+} from "nats";
 
 const BASE_DELAY_MS = 100;
 const MAX_DELAY_MS = 3_000;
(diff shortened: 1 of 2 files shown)


tlgimenes:
@cubic-dev-ai

PR #3110, opened 2026-04-13
fix(mesh): validate redirect_uri in oauth-proxy to prevent OAuth hijacking

0xcucumbersalad:
The oauth-proxy authorize endpoint was forwarding any redirect_uri to the upstream auth server without validation. An attacker could supply a malicious redirect_uri (e.g. https://evil.com) and steal authorization codes after consent.

Now the proxy requires redirect_uri and rejects any host that is not studio.decocms.com (or subdomains) or localhost/127.0.0.1.

---
## Summary by cubic
Secures the oauth-proxy authorize endpoint by validating redirect_uri (optional per RFC 6749) to prevent OAuth hijacking. If present, its origin must match getSettings().baseUrl (or the request origin); invalid values return 400 invalid_request and duplicate redirect_uri params are collapsed to a single validated value.

<sup>Written for commit 21c0be58d2d4c8bdba57aa70eed8ea57b9a4c75b. Summary will update on new commits.</sup>

### DIFF
--- a/apps/mesh/src/api/app.ts
+++ b/apps/mesh/src/api/app.ts
@@ -672,7 +672,10 @@ export async function createApp(options: CreateAppOptions = {}) {
       const redirectUri = targetUrl.searchParams.get("redirect_uri");
       if (!redirectUri) {
         return c.json(
-          { error: "invalid_request", error_description: "redirect_uri is required" },
+          {
+            error: "invalid_request",
+            error_description: "redirect_uri is required",
+          },
           400,
         );
       }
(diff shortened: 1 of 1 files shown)


apps/mesh/src/api/app.ts
```diff
@@ -668,6 +668,36 @@ export async function createApp(options: CreateAppOptions = {}) {
...
+      // Validate redirect_uri to prevent OAuth hijacking — only allow our own origins
+      const redirectUri = targetUrl.searchParams.get("redirect_uri");
+      if (redirectUri) {
+        try {
+          const redirectHost = new URL(redirectUri).hostname;
+          const allowed =
+            redirectHost === "localhost" ||
+            redirectHost === "127.0.0.1" ||
+            redirectHost === "studio.decocms.com" ||
+            redirectHost.endsWith(".studio.decocms.com");
```
viktormarinho: baseurl

viktormarinho (approved):
lgtm

PR #3159, opened 2026-04-23
fix(deco-sites): generate connId server-side and ensure member row idempotency

guitavano:
## Summary

- **Security fix**: `connId` was being generated on the client and sent in the request body, allowing any authenticated user to supply an arbitrary UUID as the connection primary key. The backend now generates `connId` server-side via `generatePrefixedId("conn")` and returns it in the response — the client uses whatever the server hands back.
- **Bug fix**: `getOrCreateTeamServiceAccount` was returning early (with just the API key) when a profile already existed, skipping the `members` and `member_roles` creation steps. The function now checks for a missing member row and creates it on the fly, making the whole flow idempotent.

## Test plan

- [ ] Import a deco.cx site — connection is created successfully, `connId` in the response matches the one used by the MCP tool call
- [ ] Trigger the import for a team whose profile exists but is missing a `members` row — membership is created automatically
- [ ] Confirm the request body no longer accepts/uses a client-supplied `connId`

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Generate connection IDs on the server and make team service account creation idempotent to close a security gap and stabilize deco site imports. The client no longer sends `connId`; it uses the ID returned by the API.

- **Bug Fixes**
  - Generate `connId` on the server via `generatePrefixedId("conn")`; ignore client-supplied IDs and require only `siteName` and `orgId`.
  - Ensure `getOrCreateTeamServiceAccount` creates missing `members` and `member_roles` rows when a profile exists, making the flow idempotent.

- **New Features**
  - Add “Pull Requests” and “Releases” to the default main view for new deco site imports.

<sup>Written for commit 7d4e097411aa7b5dbd26601942be3ace2fcf5698. Summary will update on new commits.</sup>

### DIFF
--- a/apps/mesh/src/api/routes/deco-sites.ts
+++ b/apps/mesh/src/api/routes/deco-sites.ts
@@ -13,6 +13,7 @@
 import { Hono } from "hono";
 import type { MeshContext } from "../../core/mesh-context";
 import { getUserId } from "../../core/mesh-context";
+import { generatePrefixedId } from "../../shared/utils/generate-id";
 import { fetchToolsFromMCP } from "../../tools/connection/fetch-tools";
 
 type Variables = { meshContext: MeshContext };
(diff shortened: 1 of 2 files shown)


viktormarinho (approved):
ok

PR #3243, opened 2026-04-30
feat(sandbox): user-data skill for org-scoped file access

viktormarinho:

Adds a read-only `user-data` skill to the sandbox image so resident skills (pptx, docx, ...) can act on files stored in the org's object storage. Two new mesh endpoints (`/api/sandbox/user-data/list` + `/get`) authenticate via the existing per-sandbox `DAEMON_TOKEN` — `BoundObjectStorage`'s auto-org-prefix is the security boundary, so the sandbox can never name another org's keys. Also extends the `bash` tool description so the model discovers `/mnt/skills/public/` and the canonical `user-data → format-skill` chain.

Manually verified end-to-end (list, download via `mesh-storage://`, chain into pptx).

## How to Test

1. `bun run --cwd=apps/mesh migrate` to apply migration 073.
2. `bun run dev`, attach a file in chat, then exec into the sandbox:
   - `user-data-list chat-uploads/` shows the upload.
   - `user-data-download mesh-storage://chat-uploads/<key>` lands at `/home/sandbox/<basename>`.
   - Chain into a format skill (e.g. `pptx-extract \"\$(user-data-download mesh-storage://...)\"`).

## Migration Notes

- Applies migration `073-sandbox-runner-state-token-index` (btree on `state ->> 'token'`).
- Set `STUDIO_MESH_URL` to the in-cluster mesh DNS for prod K8s deploys; Docker dev defaults to `http://host.docker.internal:3000` via host-gateway.
- Sandboxes provisioned before this PR have no `state.tenant` and will return 401 on user-data calls until reprovisioned (next `ensure()` mints fresh state).

The sandbox→user file-share path (model produces an artifact, user gets a download chip) is split into a follow-up PR — branch `viktormarinho/user-data-share`.

### DIFF
--- a/apps/mesh/migrations/073-sandbox-runner-state-token-index.ts
+++ b/apps/mesh/migrations/073-sandbox-runner-state-token-index.ts
@@ -1,0 +1,18 @@
+/**
+ * Btree index on `sandbox_runner_state.state->>'token'` so the
+ * sandbox-user-data bearer middleware can resolve a DAEMON_TOKEN to its
+ * sandbox row in O(log n) instead of scanning every active sandbox.
+ */
+
+import { sql, type Kysely } from "kysely";
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`
+    CREATE INDEX IF NOT EXISTS sandbox_runner_state_token_idx
+    ON sandbox_runner_state ((state ->> 'token'))
+  `.execute(db);
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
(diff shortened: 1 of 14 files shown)


viktormarinho:
Superseded by #3245 — pivoted to studio→sandbox tools (`copy_to_sandbox` + `share_with_user`) instead of HTTP sandbox→studio direction. Cleaner: no new auth, no migration, no token plumbing, bytes stream S3↔daemon directly via presigned URLs. Closing this.

PR #3362, opened 2026-05-14
fix(automations): include all MCP prompt-message roles in chat parts

viniciusventura29:

Automations whose stored messages were built from MCP-prompt slash-mention chips (e.g. five `/Predict And Apply` chips) were either producing empty assistant turns ("No response was generated") or, when the chips contributed no inline text, failing every fire with `No user message found in input` flooding pod logs in `deco-mcp-mesh`. The root cause is a hard `role !== "user"` filter inside `promptMessagesToParts` in `apps/mesh/src/web/components/chat/derive-parts.ts` that drops the entire body of any MCP prompt that returns `role: "assistant"` or `role: "system"` — which is the common case for instruction-style prompts. This PR removes that filter, adds a server-side invariant in `normalizeMessages` so `AUTOMATION_CREATE` / `AUTOMATION_UPDATE` reject payloads without a non-empty user message, and covers both with new colocated bun tests.

## How to Test

1. Run `bun test apps/mesh/src/web/components/chat/derive-parts.test.ts apps/mesh/src/tools/automations/normalize-messages.test.ts` — 13 tests should pass.
2. In Studio chat against any virtualMCP, send a single message whose only content is a slash-mention for an MCP prompt that returns `role: "assistant"` or `role: "system"` content; confirm the assistant produces a real response instead of "No response was generated".
3. Try creating an automation whose tiptap doc has only an empty paragraph and confirm `AUTOMATION_CREATE` now throws "Automation messages must include at least one user message with non-empty content" instead of persisting an empty payload.

## Migration Notes

No schema changes. Existing broken automations (e.g. on `vir_H62GkISGy4YowOaV06okW`) keep their corrupted `messages` until re-saved; opening each one in Studio and saving rewrites `automations.messages` from the intact `metadata.tiptapDoc` through the fixed derivation path.

## Review Checklist
- [x] No breaking changes

---
## Summary by cubic
Reverts the change that included assistant/system MCP prompt-message roles when deriving chat parts, restoring the prior user-only filter. Also rolls back the server-side check that required at least one non-empty user message.

<sup>Written for commit 1841415785a93d931eadc95eea093f4c8626912e. Summary will update on new commits.</sup>


viniciusventura29:
Fechando: o diagnóstico inicial estava errado. As automations não estão falhando pelo filtro `role !== "user"` no `derive-parts.ts` — investigação no DB mostrou que elas têm conteúdo válido (parts com corpo do prompt, role user), e os threads de fire estão sendo criados normalmente. As falhas são intermitentes (e.g. "Predict and apply" tem 277 sucessos vs 139 falhas), provavelmente timeout/rate limit/sandbox. Vou investigar o root cause real em outro branch.

PR #3377, opened 2026-05-15
feat(decopilot): pending-message inbox above the chat input

viktormarinho:
## Summary

Visible queue management on top of the per-thread DBOS gate from #3376.

**Backend**
- `queued_messages` table (migration 078) — pending rows inserted on POST, atomically claimed by the workflow on dispatch. `claim()` is replay-safe (tri-state: claimed / cancelled / missing).
- `GET /:org/decopilot/threads/:id/queue` for initial snapshot + `DELETE /:org/decopilot/threads/:id/queue/:messageId` for cancel (atomic CAS).
- `streamBuffer.publish()` for emitting `data-queue-*` events onto the per-thread JetStream subject — tail subscribers see them interleaved with run chunks.

**Frontend**
- `useThreadQueue` hook (react-query backed) — fetches initial snapshot, exposes `apply(event)` so `useThreadChat`'s `onData` can forward `data-queue-*` chunks as deltas.
- `<QueuedMessages>` renders pending bubbles above the input with an X to cancel.
- Chat input no longer disabled mid-stream — typing enqueues onto the gate. The submit button is a stop button only when the composer is empty.

Built on top of #3376 (target branch for now; rebases to `main` after the stack merges).

## Known UX gap (not in this PR)

`useThreadChat.sendMessage` optimistically adds the user message to local chat state, so while a message is queued it shows both as a chat bubble and as a pending bubble. Likely fix: suppress the optimistic add while a run is already in-flight, or hide chat history entries that are still in the queue. Worth a separate design pass before deciding.

## Test plan

- [x] `bun run check` clean across all workspaces
- [x] 504 tests pass (decopilot + dispatch-queue + automations + storage)
- [ ] Manual: queue two messages on the same thread mid-stream and confirm:
  - both render as pending bubbles
  - clicking X cancels before dispatch
  - the second dispatches after the first finishes

---
## Summary by cubic
Adds a pending-message inbox above the chat input so users can see and cancel queued messages while a run is in progress. Typing during a stream now enqueues to the per-thread gate instead of disabling the input.

- **New Features**
  - Backend: `queued_messages` table with replay-safe `claim()`; queue events published via `streamBuffer.publish` as `data-queue-*`.
  - API: `GET /:org/decopilot/threads/:threadId/queue` and `DELETE /:org/decopilot/threads/:threadId/queue/:messageId`.
  - POST response: returns `{ taskId, queuedMessageId }` for idempotency/cancel.
  - Frontend: `useThreadQueue` hook (+ `decodeQueueChunk`) to sync inbox via deltas.
  - UI: `<QueuedMessages>` renders pending bubbles with cancel.
  - Input: editor stays enabled mid-stream; send enqueues; button shows stop only when the composer is empty.

- **Migration**
  - Run DB migration `078-queued-messages`.

<sup>Written for commit f2a5fb3a5187b4d5d91ba8b9b8f50a1750fefef8. Summary will update on new commits. <a href="https://cubic.dev/pr/decocms/studio/pull/3377?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/apps/mesh/migrations/078-queued-messages.ts
+++ b/apps/mesh/migrations/078-queued-messages.ts
@@ -1,0 +1,46 @@
+/**
+ * Queued messages — pending agent runs awaiting dispatch on the per-thread
+ * gate queue.
+ *
+ * Row lifecycle:
+ * - Inserted with status='queued' at POST /messages, before the workflow is
+ *   enqueued. Holds the message text so the inbox UI can render pending
+ *   bubbles above the input.
+ * - On DELETE /messages/:id, the row is atomically transitioned to
+ *   'cancelled' (CAS). The workflow consumer checks this on dequeue and
+ *   skips dispatch if the row is gone or cancelled.
+ * - The workflow consumer deletes the row at the start of dispatch (the
+ *   message is now "running" — it appears in the chat as the streaming
+ *   response, not in the inbox).
+ *
+ * No 'running' / 'consumed' states persisted — the workflow journal is the
+ * source of truth for in-flight work, this table only tracks the inbox.
+ */
+
+import type { Kysely } from "kysely";
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable("queued_messages")
+    .addColumn("id", "text", (col) => col.primaryKey())
+    .addColumn("thread_id", "text", (col) => col.notNull())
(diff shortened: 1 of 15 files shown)


viktormarinho:
Holding off — has known bugs (double-render: queued messages show both as chat bubble and pending bubble). Will reopen after the UX is sorted.

PR #3400, opened 2026-05-19
feat(tasks-panel): infinite scroll for tasks sidebar

viniciusventura29:

The tasks sidebar previously loaded a single fixed page (50 items) and stopped — orgs with more tasks couldn't reach them from the panel. This PR turns the list into an infinite-scroll experience:

- `useTasks` switches from `useSuspenseQuery` to `useSuspenseInfiniteQuery`, paging the `COLLECTION_THREADS_LIST` call by `offset` in steps of `TASKS_PAGE_SIZE`.
- `TasksPanel` exposes `hasMore`/`isFetchingMore`/`onLoadMore`, combining the two parallel queries (me + automation).
- `TasksSection` mounts an `IntersectionObserver` sentinel on the last row via `useInfiniteScroll`, with the scroll container as the observer root (prevents cascade-loading short lists that already fit the viewport).
- `overscroll-contain` on the scroll container blocks scroll chaining to the page.
- Cache helpers (`cache-operations`, `read-cached-last-thread`, `read-cached-task-branch`) migrate to the new `{ pages: TasksPage[], pageParams }` shape.

## How to Test

1. Open the tasks sidebar in an org with > 50 open tasks (mix of me-owned and automation-triggered).
2. Scroll down inside the sidebar list — additional pages should load as the last row enters the scroll container.
3. Continue scrolling past the end — the page itself must not scroll (overscroll containment).
4. With < 50 tasks total, the list should render once and stop without firing extra fetches.

## Review Checklist
- [x] No breaking changes

---
## Summary by cubic
Adds infinite scroll to the tasks sidebar so orgs can browse beyond the first 50 items. Uses a container-rooted sentinel for reliable loading without scrolling the page.

- **New Features**
  - Infinite scroll for “My” and automation tasks; loads when the end-of-list sentinel enters the sidebar.
  - Observer is rooted to the panel container via `useInfiniteScroll(rootRef)`; container uses `overscroll-contain`.
  - Shows a lightweight “Loading more…” indicator while fetching.

- **Refactors**
  - `useThreads` now uses `useSuspenseInfiniteQuery` from `@tanstack/react-query` with offset-based paging; exposes `fetchNextPage`/`hasNextPage`/`isFetchingNextPage`.
  - Cache shape is paged `{ pages, pageParams }`; updated event helpers and cache readers to walk pages.
  - Infinite scroll hardened: dedicated sentinel after the list, stable ref callback with microtask-root setup and a 200px `rootMargin`; `TasksPanel` now sorts by `updated_at` desc to match server order.

<sup>Written for commit 2f5ebc4609b8b919b3fc90b9c84ad79182c2a623. Summary will update on new commits. <a href="https://cubic.dev/pr/decocms/studio/pull/3400?utm_source=github">Review in cubic</a></sup>


tlgimenes (approved):

tlgimenes:
# Review PR Summary
## Findings

### **#1 (Important) IntersectionObserver cascades and silently fetches every page**

`tasks-section.tsx:184` attaches the load‑more sentinel to the last rendered task row, and the observer in `use-infinite-scroll.ts:60` uses the default threshold of 0. The "Tasks" list in this panel defaults to `memberFilter = "mine"` (`tasks-section.tsx:67`) and additionally filters by `filter` (`all`/`manual`/`automation`). The combined effect: when the visible/filtered list is short enough to fit inside the scroll container (the common case — e.g. a user with 5 of their own threads in a 500‑thread org), the last row is already intersecting on first mount. The hook re‑creates its callback ref every render (no memoization — `use-infinite-scroll.ts:55`), so each `isFetchingNextPage` flip disconnects and re‑observes, the new observer immediately reports `isIntersecting`, and `fetchNextPage` keeps firing until the server runs out of pages. The new `rootRef` does not prevent this because the sentinel sits *inside* the scroll container's bounds. Net effect: any filtered view will eagerly fetch the entire org's threads in the background after first render, defeating the goal of paging. Realistic fixes: (a) put a dedicated sentinel `<div>` *after* the last row so it only intersects when the user has actually scrolled to the bottom, plus a positive `rootMargin`/`threshold`, (b) gate `onLoadMore` on `scrollContainer.scrollTop > 0`, or (c) push the `mine`/`hasTrigger` filters into the `where` clause server‑side so the page that comes back is what the user actually sees.

`apps/mesh/src/web/components/chat/task/thread-store.ts` (cascade is structural; the sentinel attaches in `apps/mesh/src/web/layouts/tasks-panel/tasks-section.tsx:184`)

### **#2 (Minor) `rootRef.current` is `null` on the initial observer attach**

In `use-infinite-scroll.ts:70`, the observer is constructed with `root: rootRef?.current ?? null`. React commits child refs before parent refs, so on the *first* mount the row's `lastElementRef` callback runs before `scrollContainerRef` is assigned, and the very first observer is created against the viewport (`root: null`) instead of the scroll container. Because the callback ref isn't memoized it's recreated on the next render and the observer is rebuilt with a valid `rootRef.current`, so the bug is self‑healing within a tick — but it makes the first‑intersection check meaningful against the viewport, not the container, and it's easy to introduce a regression by adding `useCallback`. Reading the ref inside a `useEffect` against `[rootRef.current]`, or — simpler — passing the resolved `HTMLElement | null` instead of a ref, makes the behavior deterministic.

`apps/mesh/src/web/hooks/use-infinite-scroll.ts:70`

### **#3 (Minor) `useThreads` typed `fetchNextPage` as `() => unknown` loses the awaitable**

`thread-store.ts:42` declares `fetchNextPage: () => unknown` on `UseThreadsResult`. React Query returns a Promise here, and several callers downstream (e.g. anyone wanting to await before re‑rendering, or to chain `.catch` for error toasts) lose that capability through this interface. Forwarding the real signature (`() => Promise<InfiniteQueryObserverResult<TasksPage>>` or at minimum `() => Promise<unknown>`) keeps the door open.

`apps/mesh/src/web/components/chat/task/thread-store.ts:42`

### **#4 (Nit) Sort key in panel diverges from server ordering**

`tasks-panel/index.tsx:45` re‑sorts the merged list by `created_at desc`, but the server returns rows ordered by `updated_at desc` (`thread-store.ts:78`). For offset pagination this is fragile: a row whose `updated_at` moved up between fetches can appear on two pages or be skipped, and the client re‑sort hides that from review. Not introduced by this PR, but it becomes more visible once we keep loading pages instead of stopping at 50. Worth a TODO if not addressed now.

`apps/mesh/src/web/layouts/tasks-panel/index.tsx:45`

PR #3529, opened 2026-05-28
fix(web-search): decouple deep research provider from chat provider

viktormarinho:

Self-hosted customer reported `web_search` failing when the chat model was routed through LiteLLM, even with a Gemini key configured on the org's `web_research` tier. Root cause: `dispatchRunAndWait` activated a single `MeshProvider` from the chat credential and reused it for every built-in tool, so when the chat provider was LiteLLM (or any provider without `asyncResearch`), `web_search` fell back to a tool-less `streamText` call and returned no grounding. `resolvePerRequestModels` already resolved the `web_research` and `image` tier credentials but discarded them in `toModelInfo`. This PR plumbs those credentialIds through `ModelsConfig` → `dispatchRunAndWait` (parallel-activates a second provider when the tier credential differs from chat) → `HarnessProcessLocal` → `BuiltinToolParams`, so `generate_image` and `web_search` use the right provider regardless of what the chat model is. When the tier shares the chat credential (or no tier is configured), the chat provider is reused — no extra activation, no behavior change for the baseline case.

## How to Test

1. In an org, configure two AI provider keys: set the chat tier to a LiteLLM key (or any non-Gemini provider) and the `web_research` tier to a Gemini key with `deep-research-pro-preview-12-2025`.
2. Open chat, ensure the chat model is the LiteLLM one, and send "Pesquise as notícias de hoje" (or any prompt that triggers `web_search`).
3. Expected: `web_search` enters the durable async-research path (a row appears in `async_research_jobs`) and the tool call returns a report with citations. Baseline (chat = Gemini, web_research = Gemini) keeps working unchanged.

## Migration Notes

No DB migration. Persisted runs from before this change keep working: `PersistedToolModelInfoSchema.credentialId` is optional, and `toModelsConfig` falls back to the chat credential when the per-tier credentialId is absent (matches the legacy single-provider behavior on resume).

## Review Checklist
- [x] Changes are tested and working (bun run check + affected unit tests pass)
- [x] No breaking changes

---
## Summary by cubic
Fixes `web_search` failing when chat uses `LiteLLM` by decoupling deep-research and image providers from chat. Also enforces role-based access to `image` and `web_research` tiers so only allowed keys/models are used.

- **Bug Fixes**
  - Preserve per-tier `credentialId` for `image`/`deepResearch` in `ModelsConfig`, routes, automations, and persisted runs.
  - Activate separate providers only when a tier’s credential differs from chat; plumb `imageProvider`/`deepResearchProvider` into the Decopilot harness and built-in tools. `generate_image`/`web_search` use them; reuse chat when credentials match.
  - Enforce permissions with `filterToolTiersByPermission`: silently drop `image`/`deepResearch` tiers not allowed for the user’s role at HTTP entry and in dispatch (resume/automation). Chat stays a hard 403 when not permitted.

- **Migration**
  - No DB changes. Old runs without per-tier credentials fall back to the chat credential on resume.

<sup>Written for commit 077e53095c4a547e3296c61dcd86d84d4f21ebe4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/3529?utm_source=github">Review in cubic</a>

### DIFF
--- a/apps/mesh/src/api/routes/decopilot/dispatch-run.ts
+++ b/apps/mesh/src/api/routes/decopilot/dispatch-run.ts
@@ -53,6 +53,7 @@ import { buildOnTitleUpdated } from "./on-title-updated";
 import {
   checkModelPermission,
   fetchModelPermissions,
+  filterToolTiersByPermission,
 } from "./model-permissions";
 import type { RunRegistry } from "./run-registry";
 import { resolveThreadStatus } from "./status";
@@ -472,6 +473,10 @@ async function prepareRun(
     // 1. Check model permissions (decopilot-only; CLI harnesses run with
     //    the user's own provider credential / local CLI binary, which is
     //    already vetted at credential-creation time).
+    //    Also filters image/deepResearch tier slots: routes.ts already
+    //    strips disallowed tiers at HTTP entry, but resume + automation
+    //    paths re-enter through dispatch-run without that gate, so this
+    //    second pass keeps the policy consistent across entry points.
     if (harnessId === "decopilot") {
       const allowedModels = await fetchModelPermissions(
         ctx.db,
@@ -488,6 +493,10 @@ async function prepareRun(
       ) {
         throw new Error("Model not allowed for your role");
       }
+      input = {
+        ...input,
+        models: filterToolTiersByPermission(allowedModels, input.models),
+      };
     }
 
     const windowSize = input.windowSize ?? DEFAULT_WINDOW_SIZE;
--- a/apps/mesh/src/api/routes/decopilot/dispatch-run.ts
+++ b/apps/mesh/src/api/routes/decopilot/dispatch-run.ts
@@ -496,22 +496,39 @@ async function prepareRun(
       throw new Error("dispatchRunAndWait: taskId is required");
     }
 
-    // 2. Load entities and create/load memory in parallel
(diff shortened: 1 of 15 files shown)


viktormarinho:
**Follow-up — multi-provider e2e coverage**

Skipped a full e2e for "chat provider A + web_search provider B" in this PR. Want to add it later, but it needs the Playwright suite to gain a reusable "mock external AI/web-research provider over HTTP" fixture first (the multi-pod e2e setup apparently already has something similar — worth lifting that pattern).

The intermediate-tier inline `.e2e.test.ts` (à la `web-search.e2e.test.ts`) would cover the `getBuiltInTools → createWebSearchTool → provider.asyncResearch` routing fine, but the actual ativação dupla in `dispatchRunAndWait` and the persistence/resume round-trip only get exercised end-to-end via Playwright with two mocked AI providers.

Current coverage:
- `filterToolTiersByPermission` — unit tests (7 cases) in `model-permissions.test.ts`
- Tool registration with the new `imageProvider` / `deepResearchProvider` params — type-checked via `registration.test.ts`, `instrument-built-ins.test.ts`, `user-ask.e2e.test.ts` fixtures
- Async-research happy/resume/failure path — existing `web-search.e2e.test.ts` (single provider)

Not blocking ship — opening this for the e2e infra work.

PR #3606, opened 2026-06-01
fix(sandbox): recover git/daemon proxy when preview handle is stale

guitavano:
## Summary
- Resurrect or adopt a live sandbox claim in `proxyDaemonRequest` before returning 404, matching the recovery path preview already uses
- Retry git/daemon proxy routes once via `adoptLiveClaim` before mapping 404 → 410 Gone
- Self-heal the Save changes / publish dialog by re-running `SANDBOX_START` once when git APIs return 410

Fixes the production issue where preview stays up but `GET /api/:org/sandbox/:virtualMcpId/:branch/git/status` returns 410 with "Sandbox handle is gone".

## Test plan
- [ ] Reproduce stale-handle case: preview iframe loads, open Save changes — git status should succeed instead of 410
- [ ] Confirm `git/status`, `git/diff`, and publish still work on a healthy sandbox
- [ ] Confirm truly deleted sandboxes still return 410 and trigger reprovision
- [ ] Deploy to studio.decocms.com and verify the reported URL no longer 410s while preview is live

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Recover git/daemon proxy when a sandbox handle is stale so Save changes and Publish keep working while preview is live. The UI now self-heals once before surfacing 410 Gone.

- **Bug Fixes**
  - On 404 from daemon/git proxy, try `adoptLiveClaim` and retry once; only then map 404 → 410.
  - Added `adoptLiveClaim` to the agent sandbox provider and mirrored preview resurrection in `proxyDaemonRequest`, including retries after 401 or port-forward failures.
  - `git/status` and `git/diff` pass `{ userId, projectRef }` to enable adoption and retry.
  - Publish dialog detects unreachable sandbox, runs a single `SANDBOX_START` via `useSandboxStart` (`@decocms/mesh-sdk` + `SELF_MCP_ALIAS_ID`), then retries loading changes.
  - Truly deleted sandboxes still return 410 and trigger reprovision.

<sup>Written for commit e544833a6804c8d9369c7f6220c5db67f7835645. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/3606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


guitavano:
## Architectural note

The preview gateway **cannot** be out of sync with the daemon proxy — preview traffic is served **by the daemon** (port 9000, reverse-proxying to the dev server). Both paths must resolve the **same handle** to the **same live pod**.

What was happening in prod:

- **Preview** kept working because the gateway URL (`sandboxMap.previewUrl`) still routed to a live claim/pod.
- **Git/daemon APIs** (`/git/status`, etc.) went through mesh's `proxyDaemonRequest`, which held a **stale or empty records cache** and returned 404 → 410 before trying to recover.

The handle is deterministic (`computeClaimHandle`) — it does **not** change on reprovision. The bug wasn't a "new handle vs old handle" mismatch; it was mesh losing track of the **same** handle while the gateway still reached the daemon.

This PR aligns the daemon/git proxy recovery path with what preview already does: resurrect from state-store, adopt the live K8s claim, then retry — so gateway and daemon proxy always converge on the same handle → same pod.

PR #3723, opened 2026-06-09
feat(org-fs): first-class organization filesystem (manifest + service + HTTP contract + kext-free mount serve layer)

viktormarinho:
## Summary

Foundations for a **first-class, mountable organization filesystem** over the primitive object-storage keyspace (separate from Org File Configs). The goal: every harness — decopilot, decopilot-desktop, and CLI harnesses (Claude Code, Codex) reached via `bunx decocms link` — sees the same files as **real bytes on disk**, mounted lazily and **without any kernel extension**.

Design doc + decisions live in `.context/org-filesystem-proposal.md` (gitignored).

## What's here (4 commits)

1. **`feat(org-fs)`: manifest + service + HTTP contract** — `org_fs_entry` manifest (migration 104) with a monotonic change-feed `seq`, tombstones, and partial indexes; the `OrgFs` service (`stat/listDir/read/presignRead/write/mkdir/delete/move/changes/usage/backfill`); the `/api/:org/fs/:volume/*` routes; `resolveOrgFromPath` rebinds `ctx.orgFs` to the path org; ACL **decision A** (`ORG_FS_READ`/`ORG_FS_WRITE` granted to any org member); quotas (500 MB/file hard, 10 GB/volume soft).
2. **`fix(api)`: honor `HTTPException` status in the global error handler** — the app's `onError` was flattening **every** thrown error (incl. `HTTPException`) to 500, so `files.ts`/`file-uploads.ts`/`thread-outputs.ts` silently returned 500 where they intended 401/404. Extracted a tested `handleApiError`.
3. **`feat(org-fs)`: WebDAV serve layer** (`apps/mesh/src/file-storage/mount/`) — the daemon-side path to a **kext-free** mount: `OrgFsApi`, `OrgFsClient` (HTTP client for the contract), and `createWebdavHandler` (a minimal WebDAV/1 server over the contract targeting rclone's webdav backend). The OS (or rclone) mounts this via a built-in client.
4. **`fix(org-fs)`: WebDAV MOVE RFC compliance + root mtime** — adversarial-review follow-ups (MOVE→403 on self, →502 cross-host; ISO root mtime).

## Key decisions

- **Path-addressed** manifest (stable `id` deferrable as an additive migration if inbound refs ever need it).
- **Native FUSE rejected**: macFUSE needs a System-Extension approval (bad UX) *and* `fuse-native` has no Apple-Silicon build. Transport is instead **WebDAV-internal → `rclone serve/nfsmount` → built-in OS NFS client**, which mounts with **no kext, no sudo, no System Extension**.
- **No new server endpoints needed for the mount** — a whole-file-cached lazy mount handles partial reads/writes/truncate/handles locally (rclone's `--vfs-cache-mode full`).

## Testing

- **74 tests green** across the org-fs suite: unit (path helpers, `handleApiError`) + storage-integration (real Postgres + DevObjectStorage, MinIO for recursive backfill) for the `OrgFs` service, the HTTP routes, and the full WebDAV chain (WebDAV → client → in-process mesh routes → real `OrgFs`). No mocks of our own contracts.
- **Validated end-to-end on macOS**: both rclone's webdav backend and the built-in `mount_webdav` read the server correctly and mount kext-free (`"mounted by viktor"`, no System Extension). `bun run check`, `bun run lint`, `knip` all clean.

## Deliberately out of scope (follow-up)

The **daemon `MountManager`** (Bun.serve webdav + spawn `rclone nfsmount` + lifecycle) and the **mesh-side provisioning** (mint an fs token like `mintMcpEndpoint`, push `TenantConfig.mounts`). These need a real (non-sandboxed) environment to build + e2e-validate; the recipe is captured in the proposal doc. Cluster-pod FUSE/mount is also a later, privileged-sidecar concern (pod security blocks in-pod mounts); desktop-first is the primary target.

---
## Summary by cubic
Adds a first-class, mountable organization filesystem over object storage with a manifest, service, HTTP API, and a WebDAV layer for kext-free mounts. Also fixes the global API error handler to preserve intended HTTP statuses and enforces a DB constraint on entry kinds.

- **New Features**
  - Added `org_fs_entry` manifest (migration 105) with a `seq` change feed and tombstones; `OrgFs` service supports stat/listDir/read/presignRead/write/mkdir/delete/move/changes/usage/backfill with 500 MB/file and 10 GB/volume quotas.
  - Exposed `/api/:org/fs/:volume/*` routes (path via `path` query); `resolveOrgFromPath` rebinds `ctx.orgFs`; default access grants `ORG_FS_READ`/`ORG_FS_WRITE` to org members.
  - Added WebDAV serve layer and `OrgFsClient` to mount via built-in OS clients or `rclone` (no kext, no sudo); supports OPTIONS/PROPFIND/GET/HEAD/PUT/DELETE/MKCOL/MOVE.
  - Added integration tests for the service, routes, and WebDAV with real Postgres and object storage.

- **Bug Fixes**
  - Global `onError` now honors `HTTPException` status codes; non-HTTP errors still return 500.
  - WebDAV MOVE now follows RFC 4918 (self-move → 403; cross-host → 502); root mtime returns an ISO timestamp.
  - Added a DB CHECK constraint restricting `org_fs_entry.kind` to `file` or `dir`.
  - MOVE/rename and API hardening: directory moves now recreate empty subdirs; same-volume rename skips the soft volume quota; `changes` rejects non-numeric cursors with 400; reviving a deleted path as a directory clears stale hash/size; `/move` enforces a 64 KB body limit.

<sup>Written for commit ad8cfbab27bb258e85bdf8add78ae3a1084ef824. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/3723?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/mesh/migrations/105-org-fs.ts
```diff
@@ -0,0 +1,79 @@
...
+    .createTable("org_fs_entry")
+    .addColumn("organization_id", "text", (col) =>
+      col.notNull().references("organization.id").onDelete("cascade"),
+    )
+    .addColumn("volume", "text", (col) => col.notNull())
+    // Normalized path, no leading/trailing slash. "" is the volume root.
+    .addColumn("path", "text", (col) => col.notNull())
+    // Parent directory path ("" for top-level entries). Drives listDir.
+    .addColumn("parent", "text", (col) => col.notNull())
+    .addColumn("kind", "text", (col) => col.notNull())
```
pedrofrxncx: kind        text    not null check (kind in ('file','dir')),

viktormarinho: Done in 5cb267fb0 — added `check (kind in (file, dir))`. Storage only ever writes those two kinds, so it just enforces the existing invariant at the DB layer. Propagated through the stack (#3726, #3730) so the migration stays consistent.

PR #3759, opened 2026-06-10
feat(org-fs): cluster sandbox mounts — privileged sidecar + post-bind config relay

viktormarinho:
## Summary

Stage 2 of the org-fs rollout: hosted (agent-sandbox) pods get the same `org/` mounts desktop links have — **one PR, fully dormant** until the sandbox-env chart's `orgFs.enabled` is flipped (default `false`; default render has zero orgfs artifacts).

### How it works
```
claim binds → runner POSTs /_sandbox/orgfs-config (daemon token)
           → daemon relays JSON onto shared /run/orgfs (control emptyDir)
           → privileged sidecar (idle since pod boot) mounts skills+outputs
             via the SAME MountManager/mounter/invalidator the desktop uses
           → Bidirectional→HostToContainer propagation surfaces them in the
             unprivileged agent container at /app/org
```

### Key decisions
- **Delivery**: dedicated post-bind endpoint, not `claim.spec.env` (warm-pool claims reject it — resolving the Stage-1 spike's "DECISION NEEDED") and not TenantConfig (an orgFs-only patch classifies as no-op and is silently dropped by the config store). Mirrors the established DAEMON_TOKEN warm-pool pattern.
- **No cross-org leak**: warm Sandboxes are adopted once and deleted with the claim; pods are never reused across orgs.
- **`hostUsers` trade-off is explicit**: privileged + Bidirectional are incompatible with userns pods, so the template `fail`s loudly when `orgFs.enabled` is set without `hostUsers=true`.
- **Image**: `ghcr.io/decocms/studio/orgfs-sidecar` (bun + rclone + fuse3 + `user_allow_other`), published by `release-orgfs-sidecar.yaml` — exact netinit pattern, tag pinned in chart values.
- **Mesh gate**: `ORGFS_CLUSTER_MOUNTS` setting controls token minting for hosted pods (no wasted 7-day keys in clusters without the sidecar).
- Sidecar status file feeds the daemon's per-run `org/output` link gate, so the share-files-back flow (#3750) works in clusters too.

### Bug found & fixed by the e2e
#3753's read push-down **propagated** network errors when the presigned URL host is unreachable from the reader's network (dev MinIO presigns `localhost:9000` — inside a pod that's the pod itself), making rclone retry-loop instead of reading. `OrgFsClient.readResponse` now falls back to the buffered-through-mesh path (+ unit test with a real refused connection). Prod S3/R2 presigns public https, so push-down stays active there.

## Verification
**Kind e2e** (real image from the production Dockerfile, live dev mesh + MinIO, pod mirroring the chart render):
| flow | result |
|---|---|
| sidecar waits → relayed config → mounts both volumes | ✓ status file written |
| propagation into unprivileged container (uid 1000, caps dropped) | ✓ `allow_other` mounts visible |
| read volume content through mount | 298ms ✓ |
| external API add → visible in agent container | ~300ms ✓ |
| agent write → skills / outputs cloud state | ~5.3s ✓ each |

Also: 394 sandbox tests pass (8 new), all-workspace typecheck green, `helm template` verified (enabled render + guard failure + dormant default), Docker image builds + smoke-tests locally. Full notes in `.context/orgfs-stage2-e2e/RESULT.md` (untracked).

## Rollout (after merge)
1. CI publishes the sidecar image (workflow triggers on this merge)
2. sandbox-env: `orgFs.enabled=true` + `hostUsers=true` on a staging env
3. studio: `ORGFS_CLUSTER_MOUNTS=1`
4. EKS PodSecurity must permit the privileged sidecar

---
## Summary by cubic
Adds org filesystem mounts to hosted `agent-sandbox` pods using a privileged sidecar with a post-bind config relay. Ships off by default behind the `sandbox-env` `orgFs.enabled` flag and mesh `ORGFS_CLUSTER_MOUNTS`, and also adds a safe read fallback when presigned URLs are unreachable.

- **New Features**
  - Privileged `orgfs-sidecar` FUSE-mounts org volumes at `/app/org` via `rclone --allow-other`; Bidirectional propagation exposes mounts to the main container.
  - New daemon route `POST /_sandbox/orgfs-config` relays the JSON config to `/run/orgfs`; sidecar waits for it, mounts, and writes a status file used to gate the per-run `org/output` link.
  - Helm: opt-in sidecar in `sandbox-template.yaml`; render fails if `hostUsers=false` (privileged + Bidirectional require it).
  - Mesh: new setting `orgFsClusterMounts` (`ORGFS_CLUSTER_MOUNTS`) mints tokens for hosted pods; runner posts the config post-bind.
  - CI: `release-orgfs-sidecar.yaml` builds/signs/pushes the `orgfs-sidecar` image from chart-pinned tag.
  - Tooling: mark `packages/sandbox/daemon/org-fs/sidecar-main.ts` as a knip entry; `deploy/helm/sandbox-operator/vendor.sh` injects controller `podLabels`/`podAnnotations` hooks to prevent drift.

- **Migration**
  - In `sandbox-env` values: set `orgFs.enabled=true` and `hostUsers=true`.
  - Set `ORGFS_CLUSTER_MOUNTS=1` in the studio/mesh deployment.
  - Ensure cluster policy allows privileged containers and Bidirectional mount propagation, then deploy.

<sup>Written for commit c03219a903aa3b56c35c60547bcbd9a1a7248181. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/3759?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release-orgfs-sidecar.yaml
+++ b/.github/workflows/release-orgfs-sidecar.yaml
@@ -1,0 +1,113 @@
+name: Release orgfs-sidecar image
+
+# Tag is read back from deploy/helm/sandbox-env/values.yaml (`orgFs.image.tag`)
+# so the chart and the published image can't drift (netinit pattern).
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "packages/sandbox/daemon/org-fs/**"
+      - "apps/mesh/src/file-storage/mount/**"
+      - "packages/std/**"
+      - "deploy/helm/sandbox-env/orgfs-sidecar/**"
+      - "deploy/helm/sandbox-env/values.yaml"
+      - ".github/workflows/release-orgfs-sidecar.yaml"
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}/orgfs-sidecar
+
+jobs:
+  build-push:
+    name: Build & Push orgfs-sidecar image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+      attestations: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Read image tag from chart values
+        id: version
+        run: |
+          set -euo pipefail
+          VERSION=$(yq -r '.orgFs.image.tag' deploy/helm/sandbox-env/values.yaml)
+          if [ -z "${VERSION}" ] || [ "${VERSION}" = "null" ]; then
+            echo "::error::orgFs.image.tag missing from deploy/helm/sandbox-env/values.yaml"
+            exit 1
+          fi
+          echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels)
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ steps.version.outputs.version }}
+            type=raw,value=latest,enable=${{ github.event_name == 'workflow_dispatch' }}
+            type=sha,format=short
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push Docker image
+        id: build
+        uses: docker/build-push-action@v5
+        with:
+          # Repo-root context: the image bundles TS from packages/ + apps/mesh.
+          context: .
+          file: ./deploy/helm/sandbox-env/orgfs-sidecar/Dockerfile
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          platforms: linux/amd64,linux/arm64
+
+      # Catch a Dockerfile regression that would otherwise only surface as a
+      # CrashLoopBackOff sidecar in prod.
+      - name: Smoke-test built image
+        env:
+          IMAGE: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.version.outputs.version }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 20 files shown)


viktormarinho:
## Warm-pool e2e on kind — full production chain verified

Ran the real warm-pool rig locally (sandbox-operator + sandbox-env with `warmPool.size=1` + `orgFs.enabled=true`, `studio-sandbox:local`, mesh as agent-sandbox provider with `ORGFS_CLUSTER_MOUNTS=1`). This replaces the earlier simulated relay with the REAL `SANDBOX_START → mint → runner postOrgFsConfig → daemon relay → sidecar mount` chain on a genuinely warm-adopted pod.

### Warm-pool semantics (baseline, no org-fs)
| behavior | result |
|---|---|
| pool pre-warms before any claim | ✓ |
| warm adoption (`SANDBOX_START` total) | **324–352ms** |
| cold start (same image already on node) | 2,985ms — ~9x slower |
| claim `spec.env` | empty ✓; explicit env probe rejected: *"custom environment variables are not supported when using a warm pool"* |
| sentinel token | 200 on unclaimed pool pod → **401 after adoption** (rotation) ✓ |
| single-use pods | claim deleted → adopted Sandbox+pod deleted (no recycling) ✓ |
| pool replenishment | fresh Sandbox within seconds of adoption ✓ |
| contention (2 ensures, pool=1) | 1 warm adopt (324ms) + 1 cold (2,985ms) ✓ |

### org-fs on warm pods (this PR's chain, end-to-end)
| step | result |
|---|---|
| warm adoption with sidecar template | 295ms |
| runner → `POST /_sandbox/orgfs-config` (real minted token) | ✓ relayed to `/run/orgfs/config.json` |
| sidecar mounts skills + outputs, writes status file | ✓ |
| propagation into the sandbox container | ✓ real volume content listed |
| external API add → visible in pod | **892ms** |
| pod write → outputs volume cloud | 5.3s |

### Deployment finding (added to rollout notes)
The sidecar shares the pod netns, so **netinit's egress policy applies to it**: `blockCIDRs` REJECTs run *before* `allowedTCPPorts`, so the mesh URL must resolve outside the blocked CIDRs and its port must be allowed. Prod (public https URL, 443 allowed) is fine by default; kind needs `--set netinit.allowedTCPPorts={53,443,3000}` and dropping `192.168.0.0/16` from blockCIDRs to reach the host gateway. Symptom otherwise: mounts attach but every read is EIO + the invalidator logs ConnectionRefused.

PR #3897, opened 2026-06-13
feat(telos): add @decocms/telos goal-pursuit agent core

pedrofrxncx:
## Summary

Adds **`@decocms/telos`** — a small, standalone, domain-agnostic core for agents that **pursue a fixed goal**. One agent per tenant observes a world, measures it against a goal it did not choose, and acts to close the gap until none is left.

The design keeps the goal and the striving rigorously separate, so an agent can never quietly rewrite its own purpose (the names come from Aristotle's two movers — see the package README for the full intro):

- **`UnmovedMover<T>`** — the goal: frozen, versioned, per-tenant target data. Behaviorless; never mutated, never acts.
- **`GoalLedger<T>`** — append-only succession of goal versions (installing a new version is the only way a goal changes).
- **`Eudaimon<S,T,G>`** — the agent: observes, measures, deliberates, acts — pursuing the current goal but never authoring it.
- **`Domain<S,T,G>`** — the one extension point: how to perceive a world, judge it, and act in it.
- **`Deliberator`** — swappable reasoning: `ruleDeliberator` (offline, deterministic) or `aiDeliberator`.

## What's included

- `core.ts` — `UnmovedMover`, the `GoalLedger`/`Domain`/`Action`/`Deliberator` ports, `EventBus`, `Eudaimon`, `wire()`. **Depends only on `zod`** — never the AI SDK.
- `InMemoryGoalLedger` + `inMemoryBus` — swappable defaults.
- `ruleDeliberator` (offline) and `aiDeliberator` (exported at `@decocms/telos/ai`; `ai` is an **optional peer dep**, model is **injected**).
- Runnable `examples/` (storefront + content-calendar on the same core) and co-located unit tests.

## Architecture: two causes, kept apart

```
the world moved        →  state.changed  →  Eudaimon.pursue()     (the agent reacts)
an authority decides   →  goal.updated   →  GoalLedger.install()   (a new fixed star)
```

The agent reacts to `state.changed`; a new goal arrives only via `goal.updated`, which only an authority emits. The agent is never on the goal-setting path.

## Scope / impact

**Additive and inert.** Nothing in the codebase imports it yet — the existing agent loop and `apps/mesh` are untouched. The package is self-contained and extractable to its own repo (deps: `zod`; optional peer: `ai`).

## Testing

- `bun test packages/telos` — 10 pass (frozen mover, ledger succession, pursuit, causal separation, deterministic convergence to a fixed star + re-pursuit of a raised goal).
- `bun run --cwd packages/telos check` (tsc), `bun run lint`, `bun run knip` — all green.
- `bun run --cwd packages/telos example` — converges as expected (`ledger history: v1 -> v2`).

`knip.jsonc` gains one entry marking `ai` as the intentional optional peer (mirrors the existing `mesh-sdk` → `sonner` precedent).

---
## Summary by cubic
Adds `@decocms/telos`, a small, domain‑agnostic core for fixed‑goal agents, and wires it into Mesh with a Postgres goal ledger, durable onboarding, a recurring pursuit loop, and a live home card. On signup we install a deterministic, fixed org goal and stream research, thoughts, and suggestions; Decopilot and user agents carry a telos (charter + guard) resolved from per‑agent goals, and pursuit now runs continuously for the enduring goal.

- **New Features**
  - Core: `UnmovedMover` (adds `source`), `GoalLedger` (adds `anchor()`), `Eudaimon`, `Domain`/`Action`, `Deliberator` (rule or optional `@decocms/telos/ai`), `Telos<T>` (charter/guard/measure for existing agents), `applyAction`, `VetoError`; bus adds `eudaimon.action.vetoed` and `PursuitContext.vetoed()`.
  - Durable runtime + Mesh: `bootTelos()` migrates/builds the `@decocms/telos/postgres` store and registers capabilities; shared, rate‑limited DBOS queue via `initTelosDbos()`; onboarding installs a fixed Goal (`buildStudioGoal()`) on `user.signup` and runs background research (Firecrawl + OpenRouter) to propose facts; pursuit runs immediately and on a schedule with a heartbeat and never “finishes”; user‑only actions surface as `goal.suggestion` and cache per org; live thoughts stream via SSE (`goal.thought`) and cache per org; fact edits debounce a goal re‑fit; `orgActivity` pub/sub nudges pursuit.
  - UI + API: home `TelosGoalCard` shows suggested next steps, live thoughts, per‑tool connection status, and one‑click install buttons (`use-install-app`); `GET /api/:org/telos-goal` returns `{ goal, facts, suggestion, thought, progress, status }`; `POST /api/:org/telos-facts/:id` confirms/rejects facts; SSE notifications (`telos.goal.installed`, `telos.facts.updated`, `telos.goal.suggestion`, `telos.goal.thought`).
  - Agents: Decopilot resolves an agent telos from a per‑agent goal (tenant `agent:<id>`); the charter is added to the system prompt and tools are guarded via `guardTools` from `@decocms/telos/daimonion`; new agents seed their telos from their instructions on create.

- **Refactors**
  - Telos storage is self‑managed: replaced Mesh migrations/adapters with `bootTelos()`, removed legacy onboarding ledger/facts wiring, added a telos SSE pool and client hooks, and updated `knip.jsonc` to ignore the optional `ai` peer.
  - Removed the previous catalog module; tool targets come from the fixed Goal and registry data (targets now support optional `appName` and `icon`). Improved `guardTools` types and cleaned telos agent imports; expanded `packages/telos/README.md` with an integration guide. Removed the old goal‑progression capability; pursuit is continuous against the enduring goal.

<sup>Written for commit b9be9c125a092c549f7eb0a9df255d3f5f52a023. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/3897?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/mesh/src/telos/durable/pursuit.ts
```diff
@@ -0,0 +1,302 @@
...
+
+// The agent's latest recommended next step for an org — captured from the cycle's
+// suggestion so the UI can surface it (SSE is ephemeral; this survives a reload
+// within the pod). Single-pod cache; persist if it must cross pods.
+export interface PursuitSuggestion {
+  kind: string;
+  reason?: string;
+  version: number;
+}
+const latestSuggestion = new Map<string, PursuitSuggestion>();
```
vibe-dex, line 24: 🔴 **Blocker (B1): in-process `Map`s break under multiple pods.** `latestSuggestion` (here), `lastObserved` (L36, the diff-gate), and `latestThought` (`thought.ts:13`) are module-level maps, but `runPursuitCycle` runs in a DBOS queue worker on *any* pod and `GET /telos-goal` reads them on *whatever* pod the LB picked.

- `lastObserved` per-pod ⇒ pod B never sees pod A's observation ⇒ it treats every org as "new" and fires a **full LLM deliberation every cycle**, defeating the gate's purpose. Cost scales ~linearly with replica count.
- `latestSuggestion`/`latestThought` ⇒ the GET routinely returns `null` even though a cycle just produced a value (on another pod).

The "single-pod cache" comments acknowledge this, but the system is explicitly multi-pod (NATS SSE, DBOS queue). **Fix:** persist this state in Postgres (e.g. a `telos.pursuit_state` row per tenant). The diff-gate especially must be durable or it provides zero protection in prod.

apps/mesh/src/telos/durable/thought.ts
```diff
@@ -0,0 +1,26 @@
...
+// pursuing. Pushed over SSE the instant it's produced, and cached so a reload
+// mid-stream still shows the last one (the GET surfaces it). Single-pod cache,
+// ephemeral by design: a lost thought only costs a line of trace, never data.
+export interface TelosThought {
+  text: string;
+  phase: "research" | "pursuit";
+  version?: number;
+}
+
+const latestThought = new Map<string, TelosThought>();
```
vibe-dex, line 13: 🔴 **Blocker (B1):** `latestThought` is an in-process map read by `GET /telos-goal`. Under multiple pods the GET hits a different pod than the one that produced the thought ⇒ user sees `thought: null`. Persist alongside the pursuit-state fix (see comment on `pursuit.ts:24`).

packages/telos/src/storage/postgres.ts
```diff
@@ -0,0 +1,260 @@
...
+          version: r.version,
+          target: asTarget<T>(r.target),
+          source: r.source as GoalSource,
+        }),
+    );
+  };
+
+  return {
+    async install(tenant, target, source: GoalSource = "authority") {
+      const version = (await movers(tenant)).length + 1;
```
vibe-dex, line 158: 🔴 **Blocker (B2): non-atomic version computation → race + unique-constraint crash.** `version = (await movers(tenant)).length + 1` then a separate insert, no transaction, no `ON CONFLICT`. Two concurrent installs for the same tenant read the same length, both insert `N+1`, and `goals_tenant_version_idx` rejects the loser with an opaque error.

Real racing paths: `installAgentGoal` runs on **every** virtual-MCP create with no serialization (and swallows the error → silent fail), the safety-net cron + `goal.installed` re-fires, and multi-pod.

**Fix:** `version = (SELECT COALESCE(MAX(version),0)+1 FROM goals WHERE tenant=?)` in a single INSERT statement (atomic + O(1)), or a transaction with retry-on-unique-violation. (Also: this loads the entire version history just to count — see L172.)

packages/telos/src/storage/postgres.ts
```diff
@@ -0,0 +1,260 @@
...
+          version,
+          source,
+          target: JSON.stringify(target),
+        })
+        .execute();
+      return new UnmovedMover<T>({ tenant, version, target, source });
+    },
+
+    async latest(tenant) {
+      const last = (await movers(tenant)).at(-1);
```
vibe-dex, line 173: 🔴 **Blocker (B3): hot-path reads load the full version history.** `movers()` does `SELECT * … ORDER BY version ASC` (all rows); `latest()` (here), `anchor()` (L178), and `install()` (L157) all funnel through it. `latest()` runs every pursuit cycle and every GET, and engine-authored progression goals append a row per progression ⇒ unbounded per-tenant growth.

**Fix:** `latest()` → `ORDER BY version DESC LIMIT 1`; `anchor()` → `WHERE source='authority' ORDER BY version DESC LIMIT 1`.

apps/mesh/src/telos/research.ts
```diff
@@ -0,0 +1,368 @@
...
+
+export interface ResearchResult {
+  facts: FactInput[];
+}
+
+// Mocked research subject — used only as the last-resort fallback when a signup
+// carries no email. Override the whole subject with TELOS_RESEARCH_EMAIL /
+// TELOS_RESEARCH_NAME for local testing.
+export const RESEARCH_EMAIL =
+  process.env.TELOS_RESEARCH_EMAIL ?? "pedrofrxncx@deco.cx";
```
vibe-dex, line 30: 🟡 **Important (I1): hardcoded personal email as research fallback.** On a signup with no email, the system researches and persists facts about **a real, specific person** (`pedrofrxncx@deco.cx`) against an unrelated org — a privacy/data-hygiene problem. **Fix:** skip research entirely when no email is present; never default to a human's address.

vibe-dex (commented):
## Review: telos goal-pursuit core

The `@decocms/telos` kernel (`core`/`extensions`/`deliberators`) is well-built — clean layering, report-only design, exemplary pure unit tests, no hand-rolled backoff, no React-19 rule violations. **Every blocker is concentrated in the mesh durable host glue and the Postgres adapter.** Inline comments below, grouped by severity in their bodies.

### 🔴 Blocker — no tests on the production-critical host layer
`apps/mesh/src/telos/` (~20 files) and `packages/telos/src/storage/postgres.ts` have **zero tests**; no e2e specs. The *tested* deliberator path (`ruleDeliberator`) is the fallback — the *production* path (`aiDeliberator`) is untested. Before merge, at minimum: (1) `postgres.integration.test.ts` covering version monotonicity + the concurrent-install race (see comment on `install`) + jsonb round-trip; (2) co-located unit tests for the pure pursuit helpers (`sigOf`, `clampDelay`, diff-gate) and `research.ts`'s cited-or-dropped filter (`hostOf` + host allowlist — the README's confabulation backstop); (3) one e2e: signup → goal install → pursuit cycle produces a suggestion.

### ✅ Confirmed clean (non-issues)
SQL injection (fully Kysely-parameterized); SSE cross-org isolation (org-scoped, ref-counted, capped); package-vs-mesh "duplication" (it's interface→implementation, not dup); hand-rolled retry/sleep (delegated to DBOS + `Debouncer`); React-19 banned hooks (uses `useSyncExternalStore`).

**Verdict: Needs changes.**

apps/mesh/src/telos/durable/pursuit.ts
```diff
@@ -0,0 +1,302 @@
...
+// silently stop being pursued. This re-arms the basic loop for every org with a
+// goal on a heartbeat — cheap insurance that every goal keeps striving. The tick
+// diff-gates, so reviving a quiet org is nearly free (observe, no model call).
+async function pursuitSafetyNetFn(
+  _scheduledTime: Date,
+  _startedAt: Date,
+): Promise<void> {
+  const runtime = requireTelosRuntime();
+  const tenants = await DBOS.runStep(
+    () => Promise.resolve(runtime.store.ledger.tenants()),
```
vibe-dex, line 230: 🟡 **Important (I2): agent-purpose tenants leak into onboarding pursuit + no pagination.** `installAgentGoal` writes `agent:<id>` tenants into the **same** `telos.goals` ledger, and this safety net sweeps `ledger.tenants()` (`SELECT DISTINCT tenant`, full scan) and arms `runPursuitCycle` for each. So every created agent gets a daily onboarding cycle against `organization_id = 'agent:<uuid>'` (empty connections → wasted LLM-eligible cycle + SSE emit to a dead channel), and the serial fan-out over an ever-growing table won't scale.

**Fix:** filter out `agent:`-prefixed tenants here (or give the ledger a tenant-kind scope / separate store), and paginate `tenants()` (keyset).

apps/mesh/src/telos/agents.ts
```diff
@@ -0,0 +1,45 @@
...
+  await agentGoalLedger().install(agentTenant(agentId), goal, "authority");
+}
+
+// The agent's current purpose, or null if none was ever installed (older agents
+// degrade gracefully — no telos, no charter block, no guard).
+export async function latestAgentGoal(
+  agentId: string,
+): Promise<AgentGoal | null> {
+  try {
+    return (await agentGoalLedger().latest(agentTenant(agentId))).target;
```
vibe-dex, line 41: 🟡 **Important (I7): `latest()` error-swallowing strips an agent's charter on transient failures.** `try { …latest() } catch { return null }` treats *any* thrown error — including a transient DB blip — as "no goal installed", so a live agent silently loses its purpose/guardrails mid-incident rather than erroring. (Same pattern in `pursuit.ts`.)

**Fix:** have the ledger return `null` for not-found (add a `tryLatest`) and rethrow real errors.

*(Related: the `agent:<id>` tenant written here is what leaks into the onboarding safety net — see comment on `pursuit.ts:230`.)*

apps/mesh/src/telos/durable/pursuit.ts
```diff
@@ -0,0 +1,302 @@
...
+async function pursuitTickFn(organizationId: string): Promise<void> {
+  const { noGoal, changed, sleepMs } = await DBOS.runStep(
+    () => runPursuitCycle(organizationId),
+    { name: "telos.pursuit:cycle" },
+  );
+  // No goal → nothing to pursue; don't re-arm.
+  if (noGoal) return;
+  // Stay responsive while the world is moving (the agent's pace, or MIN), back off
+  // to MAX when idle. Activity pulls (pullPursuit) cut below MIN when the user acts.
+  const delay = changed ? (sleepMs ?? PURSUIT_MIN_MS) : PURSUIT_MAX_MS;
```
vibe-dex, line 216: 🟡 **Important (I5): uncapped AI deliberation driven by per-activity pulls.** `pullPursuit` is wired to every org-activity event and cuts below the MIN floor (5s debounce). Because the goal is deliberately never `satisfied()`, a chatty org fires a fresh (potentially multi-step) AI deliberation every ~5s, gated only by the global 30/60s queue limit — so one busy org can starve the shared telos queue for all orgs. **Fix:** set an explicit `maxSteps`/`maxActionsPerCycle` on the deliberator and add a per-org budget/rate limit, not just the global one.

apps/mesh/src/telos/capabilities/onboarding-research.ts
```diff
@@ -0,0 +1,65 @@
...
+    // with backoff; journaled, so a success never re-charges.
+    const result = await step(
+      "research",
+      () =>
+        researchUser({ email: event.email, name: event.name }, (text) =>
+          publishThought(event.organizationId, { text, phase: "research" }),
+        ),
+      {
+        retriesAllowed: true,
+        maxAttempts: 5,
```
vibe-dex, line 51: 🟡 **Important (I4): retry amplifies an already-expensive chain.** `researchUser` makes ~6–10 sequential network/LLM round-trips (Firecrawl scrape 45s + search 30s + several Perplexity/Sonnet calls). It's wrapped in **one** DBOS step, so `maxAttempts: 5` means a late failure re-runs the *entire* chain (re-scrape, re-search, re-charge tokens) up to 5×, and a signup burst queues behind a 3-wide, ~1–2 min/item pipeline. **Fix:** journal expensive sub-calls as their own steps so retries resume mid-chain; reconsider `maxAttempts: 5` on a step this heavy.

apps/mesh/src/telos/research.ts
```diff
@@ -0,0 +1,368 @@
...
+export async function researchUser(
+  subject: ResearchSubject,
+  onThought?: OnThought,
+): Promise<ResearchResult> {
+  const apiKey = process.env.OPENROUTER_API_KEY;
+  if (!apiKey) throw new Error("OPENROUTER_API_KEY is not set");
+  const openrouter = createOpenRouter({ apiKey });
+
+  const { email, name } = subject;
+  const domain = (email.split("@")[1] ?? "").toLowerCase();
```
vibe-dex, line 251: 🟡 **Important (I6): SSRF + prompt-injection surface.** `domain = email.split("@")[1]` is scraped via Firecrawl, and the scraped content (8k chars) flows unfenced into the LLM synthesis prompt and the live-streamed `goal.thought`. Egress is via Firecrawl (not a direct pod fetch) and React auto-escapes the render path (no XSS), so blast radius is bounded/self-targeted — hence Important not Blocker. **Fix:** allowlist/denylist the email domain (reject IP literals, `localhost`, RFC1918, internal TLDs, free-mail) and wrap untrusted scraped/searched content in clearly-delimited fences in the prompt.

apps/mesh/src/web/components/chat/index.tsx
```diff
@@ -14,7 +14,6 @@ import { HIGHLIGHT_COLLAPSED_HEIGHT_PX } from "./highlight/collapsible-highlight
 import { useHighlightCount } from "./highlight/use-highlight-count";
 import { ChatInput } from "./input";
 import { MessagePair, useMessagePairs } from "./message/pair.tsx";
-import { NoAiProviderEmptyState } from "./no-ai-provider-empty-state";
 import { CreditsEmptyState } from "./credits-empty-state";
```
vibe-dex, line 17: 🟡 **Important (I3): unrelated product change in a telos PR (scope).** This PR removes `NoAiProviderEmptyState` (here + its `Chat.NoAiProviderEmptyState` assignment, and the `showProviderEmptyState` logic in `side-panel-chat.tsx` / `home-page/index.tsx`), changing what provider-less users see on home + side panel. That's not "telos goal-pursuit core." **Fix:** split into its own commit/PR, or justify the coupling explicitly in the PR description.

apps/mesh/src/web/hooks/use-install-app.ts
```diff
@@ -0,0 +1,214 @@
...
+    connectionId: WellKnownOrgMCPId.REGISTRY(org.id),
+    orgId: org.id,
+    orgSlug: org.slug,
+  });
+
+  const [status, setStatus] = useState<Status>("idle");
+  const [activeAppId, setActiveAppId] = useState<string | null>(null);
+
+  async function install(rawAppName?: string): Promise<void> {
+    console.log("[telos-install] click", { rawAppName });
```
vibe-dex, line 55: 🟢 **Minor (M1): leftover debug logging on a live path.** Five `console.log("[telos-install] …")` calls (L55, 92, 108, 119, 132) on the one-click install flow, with JSDoc that says "Logs each step." Remove or gate behind a dev flag — connection URLs/ids in console output are noise and a minor info-leak in shared/recorded sessions.

apps/mesh/src/web/hooks/use-install-app.ts
```diff
@@ -0,0 +1,214 @@
...
+                  clientId: tokenInfo.clientId,
+                  clientSecret: tokenInfo.clientSecret,
+                  tokenEndpoint: tokenInfo.tokenEndpoint,
+                }),
+              },
+            );
+            if (!response.ok) {
+              await actions.update.mutateAsync({
+                id,
+                data: { connection_token: token },
```
vibe-dex, line 180: 🟢 **Minor (M2): verify the token fallback is vaulted.** On a failed/thrown oauth-token POST, this falls back to persisting the bearer token via the generic `connection_token` update path. Confirm that path routes through the vault the same way the oauth-token endpoint does; if it stores plaintext in connection config, the fallback downgrades token security — drop it and surface the failure instead.

packages/telos/src/core/eudaimon.ts
```diff
@@ -0,0 +1,286 @@
...
+export interface PursuitAction {
+  kind: string;
+  payload?: unknown;
+}
+export interface VetoedAction extends PursuitAction {
+  reason: string;
+}
+// A subordinate goal the proposer authored this cycle: installed (engine version)
+// or rejected by approveGoal. Absent when no proposer ran or it returned null.
+export type GoalProposal<T> =
```
vibe-dex, line 37: 🟢 **Minor (M3): duplicate type name, incompatible shapes.** `GoalProposal<T>` here is `{target, version, installed}`, but `extensions/elenchus.ts:12` exports a different `GoalProposal<T>` = `{target, rationale?, citations?}`. Same name, different shape, different subpaths — a refactor hazard. Rename one (e.g. elenchus's → `ElenchusProposal`). *(Also: `./elenchus` and `./demiurge` ship as public exports but are interface-only stubs with no implementation/consumer — consider deferring them until a real consumer exists.)*

pedrofrxncx:

[studio-bot:stale-close]

PR #4174, opened 2026-06-26
fix(dev-agent): render dev views directly against co-located desktop sandbox

viktormarinho:
## Problem

The dev/live feature (#4138) renders an MCP-app agent's views against its sandbox dev server via the synthetic `dev_<id>` connection. For a **`user-desktop`** sandbox that dev server runs on a **loopback** previewUrl (`http://<handle>.localhost:<port>`, bound to `127.0.0.1` only).

But `dev_<id>` is resolved **and fetched in the cloud pod** (`resolveDevConnection` → `servesMcp` runs server-side, and the frontend's MCP client hits `/api/:org/mcp/dev_<id>` on the studio origin). **The pod cannot reach the user's loopback** → `servesMcp` fails → `resolveDevConnection` returns `null` → the view 503s ("Dev server not running").

This is the **default** path, not an edge case: `SANDBOX_START` defaults to `user-desktop` whenever the acting user's link daemon is online. (`agent-sandbox` works today because its previewUrl is public.)

### Empirically confirmed

Against a live desktop sandbox (`...localhost:50235`), running the exact `servesMcp` probe:

| Vantage | result |
|---|---|
| Host (co-located, = local dev) | `200`, `ok:true` |
| Off-loopback (= the cloud pod) | `Couldn't connect` → `false` → null → **503** |

Root cause: the desktop ingress binds `127.0.0.1` only (`user-desktop-provider.ts`), and `<handle>.localhost` resolves to loopback — so it's physically unreachable from any host but the user's. Local dev masks the bug because the pod and the sandbox share one loopback.

## Fix (clean: co-locate the fetch)

The invariant: **a loopback previewUrl may only be fetched by a process on the user's machine.** The browser/webview *is* co-located with the daemon, so for `user-desktop` sandboxes it talks to the dev server's previewUrl **directly**, bypassing the cloud proxy. The deco dev server's CORS is `*`, so the cross-origin MCP calls are allowed. This mirrors the existing website preview, which already points its `<iframe src>` straight at the loopback previewUrl.

- **`@decocms/mesh-sdk`** — `createMCPClient` / `useMCPClient` / `useMCPClientOptional` accept an optional absolute `mcpUrl` override (connect directly to that URL instead of the `/api/:org/mcp/:connectionId` route; never the self/REST client).
- **`use-main-panel-tabs`** — dev-view auto-detection points at `${previewUrl}/api/mcp` for `user-desktop`.
- **`project-app-view` (`AppViewContent`)** — the dev view renders against the same direct client.

Gated on a `dev_<id>` connection id + `sandboxProviderKind === "user-desktop"`, so **agent-sandbox** dev views keep the existing public-URL cloud route and **regular connection UIs are untouched**. Purely additive — no regression.

## Not in this PR (documented follow-up)

Dev **tools** in the desktop harness still resolve in the cloud (`handleVirtualMcpRequest` serves the daemon's `mcp.url`). Surfacing them for a desktop sandbox needs the cluster to pass the loopback dev URL to the co-located daemon, which then opens the MCP client locally — a typed `HarnessStreamInput` change across `@decocms/harness`. The cloud path already returns null gracefully for desktop sandboxes today, so this is additive, not a regression.

Also: browser-direct works for **authless** dev servers (the default). A token-gated dev app (`MCP_AUTH_TOKEN`) would need the token surfaced to the browser — same deferral as the header-auth follow-up noted in `dev-connection.ts`.

## Testing

- `bun run --cwd packages/mesh-sdk check` ✅, `bun run --cwd apps/mesh check` ✅ (0 errors)
- `bun run lint` ✅ (0 errors; 3 pre-existing warnings in untouched files)
- `bun run fmt` ✅
- Topology reproduction above run against a live desktop sandbox.
- ⚠️ The cross-origin render itself needs verification on the desktop app / staging (can't be exercised in single-host local dev, where loopback is always reachable).

---
## Summary by cubic
Render dev MCP views for desktop sandboxes by connecting the browser directly to the sandbox’s loopback preview URL, bypassing the cloud proxy. Fixes the 503 error when the pod cannot reach `http://<handle>.localhost`.

- **Bug Fixes**
  - Added `mcpUrl` override to `createMCPClient`, `useMCPClient`, and `useMCPClientOptional` in `@decocms/mesh-sdk` to allow direct MCP connections.
  - In `use-main-panel-tabs`, point dev views at `${previewUrl}/api/mcp` when `sandboxProviderKind` is `"user-desktop"`; use `{ vmEntry, previewUrl }` from a single `useSandboxLifecycle` call.
  - In `project-app-view` (`AppViewContent`), use the same direct client for `dev_<id>` connections (gated via `parseDevConnectionId`).
  - Guarded to affect only desktop sandboxes; `agent-sandbox` and regular/self connections keep existing routes.

<sup>Written for commit 1851c81bca0a219f1e871961178bd9658677781a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4174?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


viktormarinho:
## Desktop validation (against a live `user-desktop` sandbox)

Validated the mechanism the fix rests on against a real running desktop sandbox (`delta-arietis`, wrangler active), simulating the browser with `Origin: https://studio.decocms.com`. previewUrl = `http://delta-arietis-…localhost:50235` (loopback, ingress bound `127.0.0.1` only).

| # | Check | Result |
|---|---|---|
| 1 | **Current cloud path fails** (pod → loopback) | off-loopback vantage → `Couldn't connect` → `servesMcp` false → `resolveDevConnection` null → **503** |
| 2 | **CORS preflight** from the studio origin | `Access-Control-Allow-Origin: *`, `Allow-Methods: GET,HEAD,POST,OPTIONS`, `Allow-Headers: *` |
| 3 | **Private Network Access** (Chrome public→loopback) | `access-control-allow-private-network: true` ✅ — Chrome blocks public→loopback unless the server returns this on preflight; it does |
| 4 | **No `credentials:include`** in the SDK transport | default `same-origin` → no conflict with `ACAO: *` ✅ |
| 5 | **`initialize`** straight at the loopback previewUrl | connected ✅ |
| 6 | **`tools/list`** | 60 tools, **6 `ui://` view tools** (`open_dashboard`, `open_deck`, `report_get`, …) — exactly what `use-main-panel-tabs` auto-detects |
| 7 | **`resources/read`** `ui://commerce-discovery/app` | **1.4 MB of HTML** (`<!doctype html>…`) — exactly what `AppViewContent` renders into the iframe |

The exact call sequence the fix performs (`initialize` → `tools/list` → `resources/read`) succeeds **cross-origin, directly against the desktop dev server**. Every assumption holds against a real sandbox.

### Reproduction of the bug (root cause)
Same previewUrl port (`…localhost:50235`), `servesMcp` probe verbatim from `dev-connection.ts`:
- from the **host** (co-located, = local dev): `200`, `ok:true`
- from an **off-loopback vantage** (= the cloud pod): `Couldn't connect` → `false`

Cause: the desktop ingress binds `127.0.0.1` only (`user-desktop-provider.ts`) and `<handle>.localhost` resolves to loopback, so it's unreachable from any host but the user's. Local dev masks it because pod==machine share one loopback. This is the default path (`SANDBOX_START` defaults to `user-desktop` when the link daemon is online).

### Not covered here
No UI click-through with this branch serving the browser — that needs a separate rig + a desktop sandbox bound to it + browser automation (the live sandboxes belong to another rig). The mechanism (every call + header) is proven against the real dev server; residual risk is UI wiring only, covered by `tsc`/lint.

PR #4208, opened 2026-07-01
fix(dispatch): reap gate workflows orphaned by a deploy

tlgimenes:
## Summary

A cluster rollout that recycles the **worker pod mid-run** leaves the `threadGateWorkflow` driving that run **orphaned**: `dispatchRunAndWait` completes (the run finished on the desktop — daemon logs `dispatch complete`), then the executor dies before `consumeRunProjection`. DBOS's own recovery does **not** re-adopt the dead executor's `PENDING` workflow (conductor recovery isn't firing; a worker redeploy has nothing to re-drive its in-flight gates). So the workflow sits `PENDING` forever, holding the per-thread queue partition (`concurrency=1`) — every later message on that thread piles up `ENQUEUED` and **the thread is bricked** until manually cancelled.

Traced live on studio-stg (thread `1f1c620e`): executor `b0ddbf15` went silent at the exact instant a rollout replaced its pod; the gate's steps were `trackMessageStarted[ok] dispatchRunAndWait[ok]` with no `consumeRunProjection`, frozen `PENDING` for ~9 min with two real user messages stuck behind it.

### Why the existing reaper doesn't catch it

`thread-gate-reaper.ts`'s `reapStuckRunsSweep` targets stuck **runs** — it force-fails a run whose progress stalled, which unblocks a gate *waiting inside* `dispatchRunAndWait`. It can't see this case: the run already **completed**, so it isn't `in_progress`. The orphan is at the **workflow** layer, after dispatch returned.

## Fix

New companion sweep `reapOrphanedGatesSweep`: cancel any `threadGateWorkflow` still `PENDING` on the `thread-gate` queue whose `dispatchRunAndWait` step completed **more than 5 min ago** (`dbos.operation_outputs.completed_at_epoch_ms`). Cancelling frees the partition slot so the `ENQUEUED` backlog drains.

The grace keyed on the **step's own completion timestamp** is immune to false positives:
- a run still dispatching has no completed `dispatchRunAndWait` → not matched;
- one that just finished and is projecting is well inside the 5-min grace (projection is seconds) → not matched;
- a long run's executor liveness is irrelevant (per-gate signal, not thread-update or executor-heartbeat based).

Runs behind the existing `LINK_DURABLE_REAPER` flag, alongside the run-status sweep.

⚠️ **Deploy note:** `LINK_DURABLE_REAPER` is currently **unset in stg** — the whole reaper safety-net (both sweeps) is dormant there. Enable `LINK_DURABLE_REAPER=1` for this to take effect.

## Testing

- New pure unit test `thread-gate-reaper.test.ts`: sweep passes `now - grace` as the cutoff, cancels each returned orphan, returns the count; no-ops when empty. RED (missing export) → GREEN.
- `bun test apps/mesh/src/dispatch-queue/thread-gate-reaper.test.ts` → **2 pass**. (The `real Postgres` integration tests need a DB — they run in CI.)
- `bun run lint` → 0 errors; `bun run fmt` applied; typecheck: no new errors from these files.
- The SQL discriminator was validated against the live studio-stg `dbos` schema during diagnosis.

## Companion

Companion to #4204 (sandbox reaped mid-run). That one stops a **live** run from being killed; this one frees a thread **bricked after the run already finished**. Both are symptoms of cluster-side workflow fragility across deploys.

---
## Summary by cubic
Cancels orphaned `threadGateWorkflow`s left `PENDING` after worker redeploys so per-thread queues unblock and stuck messages drain. Adds a safe sweep that targets completed dispatches where the executor died before projection.

- **Bug Fixes**
  - Added `reapOrphanedGatesSweep` to cancel gates still `PENDING` with `dispatchRunAndWait` completed >5 min ago on the `thread-gate` queue.
  - Implemented cross-org scan `listOrphanedGateWorkflows` and cancellation via `DBOS.cancelWorkflows`.
  - Runs alongside the stuck-run sweep under `LINK_DURABLE_REAPER`.

- **Migration**
  - Set `LINK_DURABLE_REAPER=1` to enable the sweep (stg currently unset).

<sup>Written for commit b969f080c2896de65f44a24f56145e21b0022f83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/api/app.ts
+++ b/apps/mesh/src/api/app.ts
@@ -1568,6 +1568,11 @@ export async function createApp(options: CreateAppOptions = {}) {
       listStuckRuns: (cutoffIso) => reaperStorage.listStuckRuns(cutoffIso),
       forceFailIfInProgress: (id, orgId) =>
         reaperStorage.forceFailIfInProgress(id, orgId),
+      listOrphanedGateWorkflows: (cutoffMs) =>
+        reaperStorage.listOrphanedGateWorkflows(cutoffMs),
+      cancelGateWorkflow: async (workflowId) => {
+        await DBOS.cancelWorkflows([workflowId]);
+      },
     });
     registerThreadGateReaperWorkflow();
   }
--- a/apps/mesh/src/dispatch-queue/thread-gate-reaper.integration.test.ts
+++ b/apps/mesh/src/dispatch-queue/thread-gate-reaper.integration.test.ts
@@ -121,6 +121,9 @@ describe("reapStuckRunsSweep (real Postgres)", () => {
       listStuckRuns: (cutoffIso) => storage.listStuckRuns(cutoffIso),
       forceFailIfInProgress: (id, org) =>
         storage.forceFailIfInProgress(id, org),
+      listOrphanedGateWorkflows: (cutoffMs) =>
+        storage.listOrphanedGateWorkflows(cutoffMs),
+      cancelGateWorkflow: async () => {},
     };
   }
 
--- a/apps/mesh/src/dispatch-queue/thread-gate-reaper.test.ts
+++ b/apps/mesh/src/dispatch-queue/thread-gate-reaper.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "bun:test";
+import {
+  reapOrphanedGatesSweep,
+  type ThreadGateReaperRuntime,
+} from "./thread-gate-reaper";
+
+function fakeRuntime(
+  overrides: Partial<ThreadGateReaperRuntime> = {},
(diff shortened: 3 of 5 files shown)


tlgimenes:
Closing — setting this reaper approach aside for now per discussion.

PR #4224, opened 2026-07-01
feat(cms-sections): render inline object unions (A | B) as a branch selector + Location widget

guitavano:
## Problem

Editing a matcher rule whose Props contain an object union (e.g. the Location matcher's `includeLocations: (Location | Map)[]`) showed **every field at once** — City, Region Code, Country **and** the Map area — with no way to pick Location vs Map. The schema form's resolver merged such unions into one flat object (the `type = "object"` fallback ran `collectProps` across all branches).

## What this does

**Resolver** (`resolve-schema.ts`): detects a plain object union — all branches inline objects, no `$ref` / `__resolveType` / `type` discriminator — and emits a new `inline-union` type carrying each branch's `title`, `schema`, and any const discriminator fields (e.g. `{ name: "max-age" }`). Existing paths are untouched: block-ref, type-discriminated, loader, `allRefs`, enum, and **`allOf`** (intersections still merge).

**`InlineUnionField`**: renders a branch `<Select>`, infers the active branch from the stored value (const discriminator first, else which fields are set), and persists a plain object (never a `__resolveType`).

**`LocationField`**: the Location branch renders a cascading country → region → city picker — ISO 3166-1 alpha-2 countries (matching `cf-ipcountry`), an interactive Brazil state map for `BR`, free-text city. deco drops object-level `@format`, so the widget is triggered by the branch's `{city, regionCode, country}` shape. The Map branch renders the coordinate field via `format: "map"`.

**Non-destructive for legacy data**: the matcher runtime ANDs fields within an entry, so a legacy rule could combine `regionCode` + `coordinates`. Editing the visible branch preserves the hidden branch's fields (`preservedOtherBranchFields`); only an explicit branch switch resets the entry.

## Regression surface

Measured on a real site's `/live/_meta` (60 defs): the only unions that change from merged→selector are `anyOf`/`oneOf` of inline objects with no discriminator. `allOf` intersections, `$ref` unions, type-discriminated unions, loaders, enums and nullable single-branch unions are all unchanged. The new behavior lives behind a new `inline-union` type + dedicated renderer, so existing block-ref/discriminated flows are byte-for-byte the same.

## Tests

New unit tests: resolver `inline-union` detection (incl. const-tagged unions still carrying discriminators, and `type`-discriminated unions still taking the block-ref path); branch inference (shape + const); legacy-field preservation; Location datasets / value / shape helpers. Full `sections-editor` suite green (368+). `bun run check` and `bun run lint` clean.

## Downstream dependency

Requires the runtime to actually emit the union in `/live/_meta`. The Location matcher schema was hand-authored (merged) in the framework — fixed in **decocms/deco-start#298** (splits it into a `Location | Map` anyOf). With that deployed, this PR renders the selector.

---
## Summary by cubic
Render inline object unions as a branch selector in the CMS and add a Location widget. Also fixes Brazil map sizing and array item header labels. Requires the runtime to emit `Location | Map` as an `anyOf` of inline objects in `/live/_meta`.

- **New Features**
  - Resolver: detect inline object `anyOf`/`oneOf` with no `$ref`/discriminator and emit `inline-union` with branch `title`, `schema`, and const discriminators (string/number/boolean; single-enum-as-const supported); `allOf` still merges.
  - `InlineUnionField`: choose/auto-infer branch; switching resets to branch defaults; edits re-apply const discriminators and preserve non-active fields; stores a plain object; forwards `previewBaseUrl`, `onAddSectionItem`, `onRequestAddSection`.
  - `LocationField`: country → region → city picker (ISO 3166-1); interactive Brazil state map for `BR`; Map branch via `format: "map"`; country list includes Cloudflare `XX` and `T1`; also supports `schema.format: "location"`.

- **Bug Fixes**
  - Brazil map: fit SVG viewBox to content (+4% padding) to remove large margins.
  - Array items: when `title` is a mustache template, show the resolved item label instead of the raw template.

<sup>Written for commit 86a83dadbaa0276b18019433d247bdba601d9d13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


guitavano:
Thanks @cubic-dev-ai — went through all 6. Fixed in a5cae076c:

- **P1 (allOf treated as a union)** — real bug, good catch. `resolved.anyOf ?? resolved.allOf ?? resolved.oneOf` meant `allOf` intersections flowed into the inline-union path and rendered one branch instead of merging. Gated the inline-union path to `anyOf`/`oneOf` only; `allOf` falls through to the object-merge path. Added a regression test (`allOf of inline objects merges, does NOT become a selector`).

- **P2 (nested SchemaForm drops FieldProps)** — fixed. Now forwards `previewBaseUrl`, `onAddSectionItem`, and `onRequestAddSection` to the branch's `SchemaForm` (dropped `containerResolveType` from the list since `SchemaForm` doesn't accept it).

- **P2 (const discriminator not re-applied on edit)** — fixed. Both the LocationField and SchemaForm branch `onChange` now merge `activeBranch.discriminators` into every update, so the branch tag survives edits even if the stored value lacked it.

- **P2 (boolean/null const discriminators)** — added boolean support to `constValue` + the discriminator types (with a test). Skipped `null`: `constValue` uses `undefined` to mean \"no discriminator\", so a `null` const would be ambiguous, and a `null` literal discriminant is not a real pattern here.

- **P2 (missing XX / T1)** — added Cloudflare's non-ISO `cf-ipcountry` values `XX` (unknown) and `T1` (Tor) to the country list, with a test asserting both are present. Rules for unknown-country / Tor traffic are now expressible.

- **P3 (countries.ts header inaccurate)** — the file is generated purely from `Intl.DisplayNames`; the differences you saw are ICU/CLDR-version variance. Reworded the header to state ISO 3166-1 + the two Cloudflare specials and that `Intl.DisplayNames` output varies by ICU version.

Full sections-editor suite green (380).

PR #4263, opened 2026-07-02
fix(mcp-oauth): let external OAuth clients use aggregate/virtual MCP endpoints

AriOliv:
## Problem

External MCP clients (Claude Desktop / Claude Code, or any RFC 9728 / MCP OAuth client) can't connect to an org's **aggregate** (`/api/:org/mcp`) or **virtual-MCP** endpoints and call tools. Three independent causes:

1. **No usable OAuth authorization server advertised.**
   - The aggregate endpoint exposes no `oauth-protected-resource` metadata → `404`.
   - Virtual-MCP endpoints try to *proxy* the downstream connection's OAuth metadata, but a virtual MCP's `connection_url` is `virtual://<id>` — nothing to fetch → `502` (`protocol must be http/https/s3`).
   - Neither advertises Studio's own **Better Auth MCP authorization server** (which supports Dynamic Client Registration). The connection `oauth-proxy` only accepts Studio's own `redirect_uri`, so it can't serve an external client that brings its own callback → `invalid_request: redirect_uri is not allowed`.

2. **`WWW-Authenticate` advertises an `http://` `resource_metadata` URL** when Studio runs behind a TLS-terminating reverse proxy (Caddy/nginx forward over http). https-only clients (e.g. Claude) reject the non-https metadata URL.

3. **MCP OAuth sessions resolve the member role from the wrong org.** External clients don't send `x-org-id` / `x-org-slug`; the org is in the URL path (`/api/:org/...`). A **multi-org** member falls through to the "single membership only" guard, resolves to **no role**, loses the owner/admin bypass, and every connection tool call `403`s `Access denied to: <tool>`.

## Fix

| File | Change |
|------|--------|
| `api/app.ts` (`mcpAuth`) | Honor `X-Forwarded-Proto` when building the `resource_metadata` origin so it advertises `https://` behind a proxy. |
| `api/routes/org-scoped.ts` | Serve Better Auth protected-resource metadata at the aggregate `/api/:org/mcp/.well-known/oauth-protected-resource` (mounted before the proxy catch-all). |
| `api/routes/oauth-proxy.ts` | For `virtual://` connections, return Better Auth metadata instead of proxying a nonexistent downstream AS. |
| `core/context-factory.ts` | Derive an org-slug hint from the request path for MCP OAuth membership/role resolution. |

Net effect: an external client discovers Studio's own Better Auth AS, registers via DCR (its own `redirect_uri` accepted), logs in against Studio, and the token resolves to the caller's real role in the path org. Downstream per-user OAuth still happens lazily when a `per_user` connection's tool is first invoked. RBAC is unchanged — `owner`/`admin` bypass; non-admins still need an explicit connection grant.

## Testing

Validated end-to-end against a deployed Studio (behind Caddy TLS) with Claude Code:
- `POST /api/:org/mcp/<virtual-mcp-id>` → `401` + `WWW-Authenticate` with an **https** `resource_metadata`.
- Well-known → `200` advertising Studio's Better Auth AS (`authorization_servers: [<studio>]`).
- DCR with a `claude.ai` `redirect_uri` → accepted; `/authorize` → `302` (no `redirect_uri is not allowed`).
- After connect, a multi-org **admin** member's tool calls succeed (previously `403 Access denied`).

---
## Summary by cubic
Enable external MCP OAuth clients (e.g. Claude) to connect to org aggregate `/api/:org/mcp` and virtual MCP endpoints. Also shorten aggregated tool name prefixes so names fit the 64-char limit.

- **Bug Fixes**
  - Serve Studio `better-auth` protected-resource metadata for aggregate and `virtual://` MCP endpoints, enabling DCR and external `redirect_uri`s.
  - Use `X-Forwarded-Proto` when building `resource_metadata` so `https` is advertised behind proxies.
  - Resolve org from `/api/:org/...` path when headers are absent to fix RBAC for multi-org members.
  - Allow per-connection RFC 8707 `resource` via `connection.metadata.oauthResource`; default to `connection.connection_url`.
  - Per-user OAuth: don’t force `offline_access`; return the 401 challenge without recording failures in both lazy-client and proxy paths.
  - Aggregate tooling: replace long slug prefixes with a short, stable namespace code in tool/prompt names to stay under 64 chars.

<sup>Written for commit e51ad76b2f036d46cdae7486b99fcacd31ec302a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/api/app.ts
+++ b/apps/mesh/src/api/app.ts
@@ -1758,10 +1758,17 @@ export async function createApp(options: CreateAppOptions = {}) {
     // Require either user or API key authentication
     if (!meshContext.auth.user?.id && !meshContext.auth.apiKey?.id) {
       const url = new URL(c.req.url);
+      // Behind a TLS-terminating reverse proxy (e.g. Caddy/nginx) the request
+      // reaches us over http, so `url.origin` would advertise an http://
+      // resource_metadata URL and OAuth-capable clients that require https
+      // (e.g. Claude) reject it. Honor X-Forwarded-Proto so the advertised
+      // URL matches the public scheme.
+      const fwdProto = c.req.header("x-forwarded-proto")?.split(",")[0]?.trim();
+      const origin = fwdProto ? `${fwdProto}://${url.host}` : url.origin;
       return (c.res = new Response(null, {
         status: 401,
         headers: {
-          "WWW-Authenticate": `Bearer realm="mcp",resource_metadata="${url.origin}${url.pathname}/.well-known/oauth-protected-resource"`,
+          "WWW-Authenticate": `Bearer realm="mcp",resource_metadata="${origin}${url.pathname}/.well-known/oauth-protected-resource"`,
         },
       }));
     }
--- a/apps/mesh/src/api/app.ts
+++ b/apps/mesh/src/api/app.ts
@@ -346,6 +346,20 @@ const oauthProxyHandler: MiddlewareHandler<Env> = async (c) => {
   let originAuthServer: string | undefined;
   const connUrl = new URL(connection.connection_url);
 
+  // RFC 8707 resource indicator forwarded to the downstream authorization
+  // server on the authorize/token legs. Defaults to the connection's MCP
+  // endpoint URL — what most servers validate against (e.g. Supabase requires
+  // the exact endpoint). Some servers only accept the *origin* and reject a
+  // path-bearing resource (e.g. Pipedream returns "Invalid or unauthorized
+  // resource parameter" for ".../v2"). Allow a per-connection override via
+  // `metadata.oauthResource` for those, falling back to the connection URL.
+  const resourceOverride =
+    typeof connection.metadata?.oauthResource === "string" &&
+    connection.metadata.oauthResource.length > 0
+      ? connection.metadata.oauthResource
+      : undefined;
+  const resourceIndicator = resourceOverride ?? connection.connection_url;
+
   if (resourceRes.ok) {
     // Origin has Protected Resource Metadata - use authorization_servers from it
     const resourceData = (await resourceRes.json()) as {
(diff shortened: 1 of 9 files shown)


vibegui:
Thanks for the contribution @AriOliv ! One thing: did you forget to add an `outbound/errors.ts` file? 

Claude: 

🔴 Critical finding: this PR does not compile

The per-user-auth robustness fix (#5) references two functions and one module that don't exist anywhere on the PR branch:

- apps/mesh/src/api/routes/proxy.ts:421-422 calls isPerUserAuthorizationRequiredError(error) and renderPerUserAuthorizationRequired(error) — and proxy.ts imports neither (I read the full import block, lines 14-39; they're absent).
- apps/mesh/src/mcp-clients/lazy-client.ts:32 does import { isPerUserAuthorizationRequiredError } from "./outbound/errors" — but apps/mesh/src/mcp-clients/outbound/errors.ts does not exist on the branch (confirmed via git ls-tree and full-tree git grep).

There is no definition of either function anywhere in the branch. The author almost certainly forgot to git add a new outbound/errors.ts. Consequences:

- bun run check (typecheck) fails — undefined import + two unresolved identifiers.
- proxy.ts would throw ReferenceError at runtime if that path executed.
- bun test would fail to load the modules.

The only green check on the PR is cubic · AI code reviewer; the actual typecheck/test CI either isn't wired to this PR or hasn't reported — the PR is mergeStateStatus: BLOCKED. This must be fixed before any merge. Ask AriOliv for the missing outbound/errors.ts.

AriOliv:
Thanks for the catch @vibegui — you were right, `outbound/errors.ts` never made it into this branch. Rather than patching it here (this branch predates the `apps/mesh` → `apps/api` restructure, and the other pieces — `namespaceCode`, `metadata.oauthResource` — were already cherry-picked in dc1dca279c), I've re-implemented the whole per-user OAuth feature on the current layout, including the missing file and the circuit-breaker piece: #6023. Closing this one in favor of it.

PR #4357, opened 2026-07-07
feat(sandbox): golden node_modules cache via reflink (skip install on hit)

pedrofrxncx:
## What

The reflink "last mile" on top of the deps cache (#4352): a per-`(repo, packageManager, lockfile)` **golden node_modules** on the shared node-local hostPath. On a hit, `cp -a --reflink=always` clones it into the repo — a copy-on-write clone that's near-instant regardless of tree size and **skips `bun install` entirely**. On a miss, the normal install runs and publishes its result as the golden for next time.

Proven on prod xfs earlier this session: reflink clone of a 686MB `node_modules` took **~1s**, and it's write-safe (writing the clone left the source intact — CoW).

## Why reflink, not hardlink

Hardlinking from a shared cache would hit the same ~1s but is **unsafe** here: shared inodes + uid-1000 pods running untrusted code + bun-doesn't-re-verify (see #4356) = one repo can poison another's packages. reflink is copy-on-write, so a pod's writes never reach the shared golden. It's the fast *and* safe option.

## Safety

- **Per-repo key** (credential-stripped cloneUrl) — the same cross-repo isolation boundary the deps cache uses; repos never share a golden. Within a repo it's one trust domain (sandbox access implies repo write).
- **Same-filesystem gate** (`st_dev` match) before any reflink — else skip and install normally.
- **CoW isolation** — verified.
- **Fail-safe everywhere** — a partial/failed restore `rm`s and falls back to `bun install`; publish is background, idempotent, atomic (temp + rename). Worst case: a no-op, never a broken boot.

## Changes

- `setup/golden-cache.ts` — keying (`lockfileHash`, `goldenNodeModulesPath`), `sameFilesystem`, `tryRestoreGolden`, `publishGolden`. Uses `child_process.spawn("cp", [...])` (args array, no shell). Independent of `install.ts` so it doesn't conflict with #4356.
- `orchestrator.stepInstall` — restore before install (hit → skip + mark succeeded), publish after a successful miss (async, best-effort).

## Testing / validation status

- 11 unit tests (keying, lockfile hashing, per-PM separation, token-refresh stability, same-fs). Orchestrator suite green, typecheck clean, daemon bundle builds.
- ⚠️ **The daemon lifecycle wiring is not yet exercised end-to-end in a pod.** The reflink *mechanism* is proven on prod xfs, but the hit→skip→publish cycle in a live boot is not. **kind can't test the fast path** (ext4 → reflink fails → fallback); validating the fallback on kind and the hit path on a prod-xfs pod is the pre-merge gate. Do not merge before that.

Follow-up (not in this PR): golden GC (stale lockfile dirs accrue until node recycle, same as the deps cache).

---
## Summary by cubic
Adds a golden `node_modules` cache using reflink to skip installs on cache hits. Publishing is now gated on a healthy dev server and the cache is pruned to prevent disk growth.

- New Features
  - Per-(repo, package manager, lockfile) golden path under `DEPS_CACHE_ROOT/golden/...` keyed by a credential‑stripped repo URL.
  - Restore before install via `cp -a --reflink=always` when cache and workdir share a filesystem; on hit, skip `bun install`. `cp` exit code is authoritative; any failure falls back to a normal install.
  - Publish only after the dev server is healthy; the snapshot strips `.vite`/`.cache`, uses atomic temp+rename, and is best‑effort. After each publish, `pruneGoldens()` drops entries older than 7d and keeps the newest 5 per repo; restore touches mtime to keep active ones fresh.
  - Kill switch: off by default; enable with `GOLDEN_CACHE_ENABLED=1|true` (Helm `depsCache.golden: true`). Requires a reflink‑capable filesystem shared with `/app`; otherwise it falls back to a normal install.

<sup>Written for commit 0c69ccc24ccdfedfb24fa7ee17d85c5c7fe209e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/deploy/helm/sandbox-env/templates/sandbox-template.yaml
+++ b/deploy/helm/sandbox-env/templates/sandbox-template.yaml
@@ -214,6 +214,14 @@ spec:
             # per-repo BUN_INSTALL_CACHE_DIR under it (see setup/install.ts).
             - name: DEPS_CACHE_ROOT
               value: "/deps-cache"
+            {{- if .Values.depsCache.golden }}
+            # Golden node_modules reflink cache (opt-in, off by default — it
+            # touches the boot install path). Requires a reflink-capable
+            # filesystem shared between the cache and the workdir; falls back
+            # to a normal install otherwise. See setup/golden-cache.ts.
+            - name: GOLDEN_CACHE_ENABLED
+              value: "1"
+            {{- end }}
             {{- end }}
             - name: DAEMON_PORT
               value: "9000"
--- a/deploy/helm/sandbox-env/values.yaml
+++ b/deploy/helm/sandbox-env/values.yaml
@@ -237,18 +237,20 @@ readOnlyRootFilesystem: true
 # Only bun reads BUN_INSTALL_CACHE_DIR — npm/pnpm/yarn/deno installs are
 # unaffected.
 #
-# No hardlinking here: the hostPath cache and the pod's /app emptyDir are
-# separate mounts, so bun's link() into node_modules fails EXDEV and it
-# falls back to copyfile on every install. The win is the saved
-# download + extract, not a zero-copy link. Near-instant install needs a
-# prebaked read-only node_modules overlay — a separate design.
+# bun links into node_modules with its default backend: on a reflink- or
+# hardlink-capable filesystem that node_modules shares with the cache the
+# install is near-instant; otherwise (e.g. a hostPath on a different mount
+# than /app) it falls back to copyfile — still saving the download + extract.
 #
 # The daemon keys the cache per-repo under this root (DEPS_CACHE_ROOT →
-# BUN_INSTALL_CACHE_DIR=<root>/bun/<repo-hash>), so sandboxes of
-# different repos never share cache entries. Within one repo the shared
-# cache is still cross-tenant when the same cloneUrl is reachable at
-# different privilege levels (public/template repos): tampered entries
-# are caught by bun's lockfile integrity check, not by this isolation.
+# BUN_INSTALL_CACHE_DIR=<root>/bun/<repo-hash>), so different repos never
+# share cache entries — the sole cross-repo isolation boundary, since bun
+# does NOT re-verify cache content on install (a tampered cached file is
+# installed as-is). Same-cloneUrl sandboxes share a cache: one trust domain
+# for private repos (sandbox access implies repo write); still cross-tenant
+# for a repo reachable by multiple orgs (public/template) — key by org if
+# that matters. Do NOT widen to a shared cache without making it read-only
+# to tenant pods.
 depsCache:
   enabled: false
   # Node path backing the cache. Created root-owned by kubelet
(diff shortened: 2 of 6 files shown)


pedrofrxncx:
**Prod cross-mount reflink confirmed** — the one real open risk is closed.

The concern was that `/deps-cache` (hostPath) and `/app` (emptyDir) might be separate mounts that can't reflink across — which *is* the case on kind (EXDEV, same dev number but reflink fails across Docker Desktop's overlay bind-mounts). On prod it works:

```
cp -a --reflink=always /deps-cache/.xtest /app/.xtest_dst   → OK
/deps-cache dev=10311   /app dev=10311
```

Verified on a live `studio-sandbox-prod` pod. So the fast path (reflink → skip install) is real on the actual prod filesystem. Kind is a false negative → code falls back to normal `bun install`, no regression.

Remaining gate before merge: exercise the daemon lifecycle (hit→skip→publish) through a live prod boot.

pedrofrxncx:
**Addressed the two pre-merge risks** (kill switch + publish race):

1. **Independent off-switch.** Golden no longer activates just because `DEPS_CACHE_ROOT` is set — it now requires `GOLDEN_CACHE_ENABLED` (chart value `depsCache.golden`, default **false**). It ships dormant and can be toggled without redeploying or rolling back the daemon image. So merging + deploying does *not* turn it on; flip the flag when a target env is validated.

2. **Publish/dev-server race fixed.** `publishGolden` is now `await`ed before `stepStart` spawns the dev server, so the reflink snapshot is taken while `node_modules` is quiescent (previously it raced Vite writing `node_modules/.vite`). Miss-path only (~1s), no effect on warm boots.

Also corrected the same false "bun re-verifies cache integrity" claim in the `depsCache` values.yaml comment (bun does not; the per-repo key is the sole boundary).

**Remaining, documented as follow-ups (not blockers now that it's opt-in + dormant):**
- No golden GC — a golden ≈ a full node_modules on the node hostPath, unreaped until node recycle. Monitor disk before enabling widely.
- Sticky bad-golden — a corrupt golden for an unchanged lockfile would be reused until node recycle. Low residual risk now that publish snapshots a quiescent tree via an atomic (temp+rename, cp-exit-checked) reflink, but there's no per-golden self-heal on dev-start failure.
- Non-hoisted workspace/monorepo repos — only the install-root node_modules is cloned; untested for nested-node_modules layouts.

Still gated on a live prod-xfs boot before enabling (`depsCache.golden=true`).

pedrofrxncx:
**GC + sticky-bad-golden addressed.**

**Sticky bad golden — killed at the source.** Publishing no longer happens right after install; it's deferred to when the dev server is confirmed **healthy**. `stepInstall` records a pending publish, and the probe's `running` transition (`entry.ts`) calls `publishPendingGolden()`. A boot that never comes up never publishes → a broken-but-exit-0 install can't become a golden that later branches reuse. No reactive-invalidation / oscillation, no cross-pod poison markers. The snapshot also strips `.vite`/`.cache` (pod-local, churny), which as a bonus removes the earlier publish-vs-dev-startup race entirely.

**GC.** `pruneGoldens()` bounds the store: drop goldens past a TTL (7d), then cap per repo (newest 5), run opportunistically after each publish. Restore `touch`es the golden's mtime so an actively-used lockfile never ages out. Safe against a concurrent restore — an in-flight reflink from a pruned golden just fails → fallback install; completed CoW clones are independent of the source.

Tests: 19 unit (keying, lockfile hash, kill switch, and TTL / per-repo-cap / .tmp-skip / per-repo-independence for prune). Orchestrator suite green, typecheck clean, daemon builds.

Remaining follow-up (minor): non-hoisted workspace/monorepo repos — only the install-root `node_modules` is cloned; untested for nested layouts. Still gated behind `depsCache.golden` (default off) + a live prod-xfs boot before enabling.

PR #4363, opened 2026-07-07
feat(helm): optional in-cluster ClickHouse via official operator subcharts

viktormarinho:
## Summary

Lets self-hosted deployments (Cogna) run ClickHouse inside their cluster for the monitoring dashboard, instead of ClickHouse Cloud / DuckDB-over-GCS.

Two new **optional** dependencies in the studio chart, both from the official [ClickHouse/clickhouse-operator](https://github.com/ClickHouse/clickhouse-operator) repo (OCI, ghcr.io), pinned to the same version (0.0.6, upstream compatibility guarantee):

- **`clickhouse-operator`** (alias of `clickhouse-operator-helm`) — operator Deployment + CRDs (`ClickHouseCluster`, `KeeperCluster`, apiVersion `clickhouse.com/v1alpha1`). Webhooks need **cert-manager** (upstream default; `certManager.enabled=false` + `webhook.enabled=false` to opt out).
- **`clickhouse-cluster`** (alias of `clickhouse-cluster-helm`) — renders the CRs. `clickhouse.spec` / `keeper.spec` are passed **verbatim** to the CRD spec, so any field is customizable (settings, podTemplate, storage, ...). Used the official CR wrapper chart instead of hand-writing the CR template in the studio chart — same result, maintained upstream, version-locked with the operator.

Both flags are independent: operator can be enabled without the CR and vice-versa (pre-existing operator in the cluster).

## Namespace scoping (o ponto do watch)

The operator supports it: `controller.watchNamespaces` → `WATCH_NAMESPACE` env on the manager. **But** the upstream chart takes only a static list (no `tpl`, no downward API), so it can't be derived from the release namespace automatically. Compensation: the studio chart **fails at render time** if the operator is enabled with an empty list, or with a list that doesn't include the release namespace while the CR is enabled — the error message prints the exact namespace to use. Cluster-wide watch via the embedded operator is deliberately not expressible; install the operator chart separately for that.

## Env wiring

When the CR renders, `CLICKHOUSE_URL` is auto-derived in the ConfigMap to `http://<cr-name>-clickhouse-headless:8123` (mirrors the `NATS_URL` pattern, including the CR-name helper mirroring the subchart's naming). Precedence: `configMap.meshConfig.CLICKHOUSE_URL` skips the auto-derive; a `CLICKHOUSE_URL` in the Secret wins over the ConfigMap (envFrom order) — that's the documented home for authenticated URLs paired with `spec.settings.defaultUserPassword`.

## Notes

- `spec.keeperClusterRef` is **required** by the CRD, so a KeeperCluster always renders (1 replica / 5Gi default, matching the single-node default of 1 CH replica / 50Gi).
- `examples/values-clickhouse.yaml` = full Cogna-ready wiring: operator + single-node cluster + otel-collector `clickhouse` exporter writing `otel_logs`. The `studio_monitoring_logs` view stays the existing one-time manual step (`apps/mesh/src/monitoring/clickhouse-setup.md`).
- Subchart `.tgz`s vendored under `charts/` like nats/otel; CI renders the example in `helm-test.yml`.
- Operator is young (v0.0.6) — single-node scope is exactly its comfort zone; backups remain the installer's responsibility.

## Testing

- `helm lint` clean; default render has zero clickhouse references.
- Render matrix: example values (CR + keeper + WATCH_NAMESPACE + exporter merge all correct), cluster-only without operator, `meta.name` override follows into the URL, meshConfig override wins.
- Both validation failure paths exercised (empty list / wrong namespace).

cc @nicacioliveira

---
## Summary by cubic
Adds optional in-cluster ClickHouse to the Studio Helm chart using the official operator subcharts, so self-hosted installs can run the monitoring dashboard without ClickHouse Cloud. Docs now also clarify deploy-first/view-last and the brief, expected dashboard error until the view is created.

- **New Features**
  - Adds optional deps `clickhouse-operator` and `clickhouse-cluster` (OCI `ghcr.io/clickhouse`, pinned `0.0.6`).
  - Validates at render time that `clickhouse-operator.controller.watchNamespaces` is set and includes the release namespace.
  - Auto-derives `CLICKHOUSE_URL` to `http://<cr-name>-clickhouse-headless:8123` when the CR is enabled; `configMap.meshConfig` or Secret values override (Secret wins for auth URLs).
  - Includes `examples/values-clickhouse.yaml` with single-node ClickHouse and OTel `clickhouse` exporter writing `otel_logs`; README/NOTES updated; CI renders the example with the release namespace.
  - `KeeperCluster` is created by default (CRD requires `keeperClusterRef`), 1 replica.
  - Docs warn that scaling replicas/shards requires DDL changes (`ReplicatedMergeTree`/`Distributed`) and `create_schema: false`, and note that the dashboard may error until the `studio_monitoring_logs` view exists.

- **Migration**
  - To enable: set `clickhouse-operator.enabled=true` and provide `clickhouse-operator.controller.watchNamespaces=[<release-namespace>]`; set `clickhouse-cluster.enabled=true`.
  - If no cert-manager, set `clickhouse-operator.certManager.enabled=false` and `clickhouse-operator.webhook.enabled=false`.
  - For a cluster-wide operator, install `clickhouse-operator-helm` separately and enable only `clickhouse-cluster` here.
  - After logs flow, create the `studio_monitoring_logs` view once; brief dashboard errors before the view is created are expected and recover automatically.

<sup>Written for commit 082840a7f5fed1e57050c1ca97aa79a0e359906b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4363?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


nicacioliveira:
Just a warning about the self-hosted ClickHouse usage included here.

For the single-node case (the default: \`replicas: 1\`, \`shards: 1\`) simple tables work normally — the \`otel_logs\` table and the \`studio_monitoring_logs\` view created by the setup step are plain \`MergeTree\`, which is exactly right for one node.

But if the installer decides to use N replicas and/or N shards, the tables created here have to change accordingly, because the data has to be replicated/distributed through the cluster's distributed nature — the CR spec alone does not do it:

- **N replicas** → the tables must be \`ReplicatedMergeTree\` (or the \`Replicated\` variant of whatever engine). Replication is coordinated through Keeper (that's why a KeeperCluster is always created). With a plain \`MergeTree\`, each replica keeps its own independent copy and the data is **not** replicated between them — you get divergent nodes, not high availability.
- **N shards** → you additionally need a \`Distributed\` table on top of the local tables. Writes are spread across shards and reads fan out through the \`Distributed\` table. Without it, each shard only holds and sees its own slice of the data, and queries hitting a single node return partial results.

So scaling the cluster is a two-part change: bump the CR **and** update the DDL engines in \`apps/mesh/src/monitoring/clickhouse-setup.md\` in lockstep. If only the CR is scaled, the extra replicas/shards spin up but the data silently stays unreplicated/undistributed. Worth spelling this out in the setup doc so nobody scales the spec expecting HA or more capacity and quietly gets neither.

nicacioliveira (approved):

viktormarinho:
Documented in 010617c7b, thanks — new **"Scaling beyond a single node (replicas / shards)"** section in `apps/mesh/src/monitoring/clickhouse-setup.md` covering exactly this: `Replicated*` engines for N replicas (Keeper-coordinated; plain `MergeTree` = divergent copies), `Distributed` wrapper for N shards (else partial results), and the operational consequence for this chart's wiring — the collector exporter auto-creates `otel_logs` as plain `MergeTree`, so multi-node means `create_schema: false` + hand-created tables, with the exporter pointed at the `Distributed` table. Also added one-line warnings next to `replicas`/`shards` in `values.yaml` and `examples/values-clickhouse.yaml` pointing at that section, so nobody scales the CR expecting HA and quietly gets neither.

PR #4426, opened 2026-07-10
fix(storage): stop OOM-killing studio-worker on large thread-message parts

pedrofrxncx:
## Summary
- Traced last night's deco-studio-worker restart storm (55 restarts/24h, one pod OOMKilled 40x over ~3h) to `SqlThreadMessagePartStorage.serializeParts`: it `JSON.stringify`s every thread-message part synchronously, with no yielding and no size bound, whenever a run flushes parts (e.g. a large tool result / file read comes through).
- `serializeParts` now yields to the event loop every 25 parts during a batch instead of doing the whole flush in one synchronous pass.
- Added a 50MB per-payload circuit breaker as a last resort for truly pathological sizes only — normal large tool output (big file reads, big stdout) is still stored in full, not dropped.
- This gap has existed unguarded since the stream-of-record feature was dark-shipped (`d54b94ca9`, June 7) — not a recent regression, just not exercised hard enough to surface until now.

## Testing
- `bun test apps/mesh/src/storage/thread-message-parts.test.ts`
- `bun run --cwd=apps/mesh check`
- `bun run fmt`

---
## Summary by cubic
Prevents OOM kills and event-loop stalls in the studio worker when flushing large thread-message parts by yielding during serialization and adding a per-payload size guard that estimates size without stringifying.

- **Bug Fixes**
  - Yield to the event loop every 25 parts in serializeParts to avoid long synchronous JSON.stringify loops.
  - Add a 50MB per-payload circuit breaker that estimates size via a cheap traversal and skips JSON.stringify for oversized payloads.
  - Bound the size estimator to avoid deep-input stalls.
  - Use serializePayload for both payload and metadata fields.
  - Preserve normal large tool outputs in full; add tests for small, large, and pathological cases.

<sup>Written for commit 7fd361b9845e020b274491edb9942978637e8c37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/storage/thread-message-parts.test.ts
+++ b/apps/mesh/src/storage/thread-message-parts.test.ts
@@ -1,0 +1,24 @@
+import { describe, it, expect } from "bun:test";
+import { serializePayload } from "./thread-message-parts";
+
+describe("serializePayload", () => {
+  it("passes small payloads through as-is", () => {
+    const payload = { type: "text", text: "hello" };
+    expect(serializePayload(payload)).toBe(JSON.stringify(payload));
+  });
+
+  it("stores large-but-legitimate payloads in full (no data loss)", () => {
+    const big = { type: "tool-result", output: "x".repeat(2_000_000) };
+    expect(serializePayload(big)).toBe(JSON.stringify(big));
+  });
+
+  it("only circuit-breaks truly pathological payload sizes", () => {
+    const pathological = {
+      type: "tool-result",
+      output: "x".repeat(60_000_000),
+    };
+    const out = JSON.parse(serializePayload(pathological));
+    expect(out.truncated).toBe(true);
+    expect(out.originalBytes).toBeGreaterThan(50_000_000);
+  });
+});
--- a/apps/mesh/src/storage/thread-message-parts.test.ts
+++ b/apps/mesh/src/storage/thread-message-parts.test.ts
@@ -12,13 +12,19 @@ describe("serializePayload", () => {
     expect(serializePayload(big)).toBe(JSON.stringify(big));
   });
 
-  it("only circuit-breaks truly pathological payload sizes", () => {
+  it("circuit-breaks a single oversized string without stringifying it", () => {
     const pathological = {
       type: "tool-result",
       output: "x".repeat(60_000_000),
     };
     const out = JSON.parse(serializePayload(pathological));
     expect(out.truncated).toBe(true);
-    expect(out.originalBytes).toBeGreaterThan(50_000_000);
+  });
+
+  it("circuit-breaks many small strings that sum past the cap", () => {
+    const chunks = Array.from({ length: 1000 }, () => "x".repeat(60_000));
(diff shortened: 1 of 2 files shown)


pedrofrxncx:
## Trace that led to this fix

**Symptom (Prometheus, `kube_pod_container_status_restarts_total` / `kube_pod_container_status_last_terminated_reason`, namespace `deco-studio`):**
- 55 `studio-worker` container restarts in 24h.
- One pod (`deco-studio-worker-7b896f4965-d5nh8`) alone was `OOMKilled` 40 times over ~3h (2026-07-09 21:22 → 00:37 UTC), across several different rollouts (image versions from `3.115.1` through `4.0.4`) — not a regression from one deploy, a standing gap.
- Memory curve per crash cycle (`container_memory_working_set_bytes`, 60s resolution): ~4MB at pod start → ~300–340MB within the first minute → OOMKilled (container limit is 2Gi — the actual peak spikes past that between scrape samples).

**Root cause (VictoriaLogs, pod logs right before each kill):**

```
event-loop-profiler
Slow db pool acquire — event-loop lag (not pool): { waitMs: 283.47, idleAtStart: 13, total: 14, max: 25 }
event-loop-stall
event-loop-stall-profile
Slow query measurement — likely event-loop lag, not slow SQL: {
  durationMs: 1245.5,
  eventLoopLagMs: 869,
  sql: "select ... from \"org_fs_entry\" where \"organization_id\" = $1 and \"volume\" = $2 and \"path\" = $3 and \"deleted_at\" is null",
  pool: { total: 14, idle: 4, waiting: 0, max: 25 }
}
```

`event-loop-stall-profile` stack (from `lagMs: 604`–`850` samples on the same pod):

```
stringify
serializeParts  file:///.../cli.js:83138
appendParts     file:///.../cli.js:83138
appendBuiltRows file:///.../cli.js:83287
emitStepParts   file:///.../cli.js:83287
onStepFinish    file:///.../cli.js:83287
...
```

That maps onto `SqlThreadMessagePartStorage.serializeParts` (`thread-message-parts.ts`) → `PartEmitter.appendBuiltRows`/`emitStepParts` (`part-emitter.ts`) → `appendPartsStep` (`background-tool-workflow.ts`). `serializeParts` was `JSON.stringify`-ing every thread-message part synchronously with no yielding and no size bound — the `org_fs_entry` query right alongside it suggests a large file read flowing through as a tool result and getting fully stringified in one blocking call, spiking RSS past the container limit.

This unguarded path has existed since the stream-of-record feature was dark-shipped (`d54b94ca9`, 2026-06-07) — not introduced by a recent change, just not exercised hard enough to surface until this traffic pattern hit it.

PR #4429, opened 2026-07-10
feat(companion): server-side MCP provisioning for the local companion app

vibegui:
## What this is

The start of a **local companion desktop app** for deco: you download it, run it, and "link" by logging in — the friendly, non-terminal replacement for `bunx decocms@latest link`. Once linked you get **two-way access with one login**:

- **Outbound** — your local Claude Code / Codex become AI providers for your orgs (exactly what `deco link` does today; the app just runs that daemon with a GUI).
- **Inbound (new)** — each of your orgs is auto-mapped as an MCP server in Claude's global config (`~/.claude.json`), so your orgs are usable inside Claude "for free".

Target shell: the new Vercel **Native SDK** (Zig host + React WebView).

## What's in *this* PR

Just the **server-side provisioning endpoint** — the smallest, highest-risk piece, isolated so we can agree on the approach before the native app lands:

- `POST /api/companion/provision` (`apps/mesh/src/api/routes/companion-provision.ts`, mounted in `app.ts`)
- Authenticates with the `deco link` bearer (same dual-auth as `/api/links/me` via `resolveLinkBearer`)
- Lists the caller's orgs and mints **one Decopilot-scoped API key per org**, returning `{ studioUrl, orgs: [{ slug, url, key }] }` — exactly what the app writes into `~/.claude.json`
- Unit-tested; `tsc` clean

The native app (`apps/companion`: Zig host + `claude_config.zig` surgical merge + React frontend) is authored separately and **not** in this PR.

---

## 🧭 The auth decision I want a gut-check on

The original plan assumed the companion app could enumerate orgs and mint per-org keys itself using the `deco link` session token. **Verified against the live API that it can't:**

| Endpoint | With the `deco link` bearer |
|---|---|
| unscoped `POST /mcp/self` | authenticates, but every management tool (`ORGANIZATION_LIST`, `API_KEY_CREATE`, `LINK_CURRENT_GET`, …) → **"Access denied"** (no org bound) |
| scoped `POST /api/<org>/mcp/self` | without the `X-MCP-Session-Auth` marker → **"Invalid API key"**; with it → org tools throw an empty error (the OAuth-session principal isn't authorized as an org member via that path) |

So the `deco link` token is an **identity/tunnel credential**, not a management credential (matches `resolveLinkBearer`: it exists to resolve the bearer to a `userId`). The web app does all management via the **cookie** Better Auth session, which the desktop app doesn't have.

### Options considered

1. **Server-side provisioning endpoint keyed to the link identity** *(this PR)* — one login, thin client, no cookies. Mints keys via `auth.api.createApiKey` (the proven `seedOrgDb` pattern) and lists orgs straight from the Better Auth `member`/`organization` tables (because `auth.api.listOrganizations` is `requireHeaders: true` and always reads the *session* user — unusable from a session-less path).
2. **WebView cookie login** — app renders a Studio page; same-origin JS lists orgs + mints keys, then bridges to native to write the file. No backend, but two auth surfaces and it hands a file-write bridge to a remote origin.
3. **Provisioning inside the `deco` CLI/daemon** over the authenticated tunnel — native app stays a pure status GUI; needs CLI + tunnel work.

I went with **(1)** for the single-login story and a trivial client. Happy to switch.

## ❓ Open question (the one thing I couldn't settle by reading)

The Decopilot agent's **stored `virtual-mcp` id is `null`** (it's addressed via the synthetic `decopilot_<orgId>`). The access layer keys permission on `ctx.connectionId`, which for Decopilot resolves to the default `"self"` — so a key scoped `{ [decopilot_<orgId>]: ["*"] }` **may not authorize** tool calls on `/api/<slug>/mcp/virtual-mcp/decopilot_<orgId>`. It might need `"self"` or the underlying connection ids instead. The scope is a one-line constant in `provisionCompanionMcps`; it just needs one live request to pin down.

## @tlgimenes — how would you proceed?

Specifically:
1. **Auth approach** — is the server-side provisioning endpoint (option 1) the right call, or do you prefer the WebView-cookie or daemon-side route? Any concern with minting per-org keys off the link identity?
2. **Decopilot key scope** — what's the correct `permissions` shape for a key that should reach an org's Decopilot agent? `{ [decopilot_<orgId>]: ["*"] }`, `{ self: [...] }`, something else?
3. **Key lifecycle** — each provision mints fresh keys (an API key value is only returned once). OK to add a "revoke prior `purpose: companion` keys for (user, org) before minting" cleanup, or is there a better pattern?

---
## Summary by cubic
Adds a server-side provisioning API and a small native desktop app that syncs each org into Claude Code as an MCP server with one login. Targets the org’s self/management MCP and mints 90‑day keys scoped to `self` for owner/admin orgs, revoking any prior companion keys.

- **New Features**
  - Server: `POST /api/companion/provision` (auth via the `deco link` bearer) returns `{ studioUrl, orgs: [{ id, slug, name, url, key }], skipped: [{ id, slug, name, reason }] }`; filters archived orgs; builds self‑MCP URLs (`/api/<slug>/mcp/self`) with `getPublicUrl()` via `companionMcpUrl()`; mints 90‑day keys with permissions `{ self: ["*"] }`.
  - App: added `apps/companion` (Vercel Native SDK — Zig host + React WebView) with bridge commands `deco.status` and `deco.provision`; reads `~/deco/session.*.json`, calls the endpoint via `curl`, and surgically merges `deco-<slug>` MCP servers into `~/.claude.json` with atomic write and `.deco-bak` backup; simple UI, headless mode (`DECO_COMPANION_HEADLESS=status|provision`), and unit tests for the merge.

- **Bug Fixes**
  - Provisioning hardening: gate key minting to owner/admin orgs and return others in `skipped` (`reason: "requires-elevated-role"`); revoke the prior `companion-<slug>` key before minting; derive `studioUrl` from the public origin; isolate per‑org errors.
  - Tests: unit tests for `companionMcpUrl` and role gating; E2E verifies keys authorize the org self MCP, members are skipped, and re‑provision revokes old keys (accepts 401 or 403 for revoked keys); frontend mount‑fetch uses a lazy initializer.

<sup>Written for commit cee40d0bfd8f799b88e573af9ab4930297677b93. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4429?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:
**Member support is intentionally out of scope here — follow-up PR.**

Provisioning only mints keys for orgs where the caller is **owner/admin**. Member orgs are returned in `skipped` with reason `requires-elevated-role` (not silently dropped).

Why: the key must grant full access to the org's self/management MCP (`{ self: ["*"] }`), but a Better Auth API key is a flat capability that **ignores the owner's role** — so handing a wildcard-over-`self` key to a non-admin member would let them call every admin tool on that org (the exact privilege escalation `apikey-scope-enforcement.spec.ts` guards against). Safe member support needs a role-carrying credential, not a flat key. Deferring that to a follow-up.

pedrofrxncx (approved):

PR #4474, opened 2026-07-12
fix(sidebar): one new chat, retargeted — stop spawning empty chats per agent click

vibegui:
## Problem

Clicking back and forth between agents in the sidebar spawned a **new empty "New chat" every time** — the list filled up with duplicates. Root cause: `useNavigateToAgent` mints a fresh `taskId` on every click and the destination route eagerly creates that thread row (`useEnsureTask`).

## Fix

There is only ever **one** new chat. If you're already sitting in an empty "New chat", clicking a different agent **re-points that same thread at the new recipient** (swaps the `?virtualmcpid` the composer/run sends to) instead of creating another empty thread. It only starts a fresh chat when you're not already in one.

- `handleOpenAgent`: when the active thread is an empty "New chat", reuse its `taskId` and set the recipient to the clicked agent (`setTaskId`), else `navigateToAgent` as before.
- Active-agent highlight now derives from the URL `virtualmcpid` (falling back to the thread row), so it stays correct after an in-place retarget.
- Extracted the shared `isThreadEmpty` check (also used by the new-thread button).

The send path already targets the URL param agent (`agent: { id: capturedVirtualMcpId }`), so retargeting an unsent chat routes the first message to the right agent.

## Testing

- `tsc --noEmit` clean, `bun run lint` clean, client build passes.
- Manual: verify clicking agents A→B→A while in a new chat keeps a single "New chat" row and updates the recipient; sending lands the message with the last-selected agent.

---
## Summary by cubic
Stops creating duplicate empty "New chat" threads when switching agents. Reuses the current unsent chat, retargets its recipient, and updates the thread row’s agent icon so only one new chat exists.

- **Bug Fixes**
  - If the active thread is an empty "New chat", keep its `taskId`, retarget the recipient on agent click, and persist the new agent to the row (`setAgent` → `COLLECTION_THREADS_UPDATE`) so the sidebar icon follows; otherwise navigate as before.
  - Read active agent from the URL `virtualmcpid` (fallback to thread) to keep the sidebar highlight correct after retargeting. Uses `useSearch` from `@tanstack/react-router`.
  - Added `virtual_mcp_id` to `ThreadUpdateDataSchema` and storage/update mappings, `ThreadManagerStore.setAgent` and `useThreadActions.setAgent`, plus a shared `isThreadEmpty` check to verify unsent chats.

<sup>Written for commit 455da2cdb65aafd495846599c99c0d75f2cfab1c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/storage/threads.ts
+++ b/apps/mesh/src/storage/threads.ts
@@ -383,6 +383,9 @@ export class SqlThreadStorage implements ThreadStoragePort {
     if (data.branch !== undefined) {
       updateData.branch = data.branch;
     }
+    if (data.virtual_mcp_id !== undefined) {
+      updateData.virtual_mcp_id = data.virtual_mcp_id;
+    }
     if (data.sandbox_provider_kind !== undefined) {
       updateData.sandbox_provider_kind = data.sandbox_provider_kind;
     }
--- a/apps/mesh/src/tools/thread/schema.ts
+++ b/apps/mesh/src/tools/thread/schema.ts
@@ -151,6 +151,12 @@ export const ThreadUpdateDataSchema = z.object({
     "Full replacement of the thread's metadata object",
   ),
   branch: z.string().nullish().describe("New git branch for this thread"),
+  virtual_mcp_id: z
+    .string()
+    .optional()
+    .describe(
+      "Re-point the thread at a different agent. Only meaningful for an unsent thread (changing the recipient of a new chat); the UI restricts it accordingly.",
+    ),
 });
 
 export type ThreadUpdateData = z.infer<typeof ThreadUpdateDataSchema>;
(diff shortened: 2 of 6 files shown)


vibegui:
Consolidated into the breadcrumb PR (#4473) — keeping everything in one PR.

PR #4485, opened 2026-07-13
feat(harness): teach agents the Deco CMS content rules (.deco/blocks only)

hugo-ccabral:
## Problem

When users ask decopilot for a content change, the agent has zero guidance about which file is the source of truth. A grep for the visible text hits both `.deco/blocks/<key>.json` and the committed `.deco/blocks.gen.json` (which contains every block, so it often matches first). Editing the gen file renders correctly in the studio preview (SSR re-eval reads it) — but production builds run `generate:blocks`, which regenerates `blocks.gen.json` from `.deco/blocks/`, and the fast-deploy KV sync (`deco-sync-blocks-to-kv`) reads `.deco/blocks/*.json` directly. The edit is silently discarded on deploy: **persisted in git, never reflected in production.**

## Fix

Append a `DECO_CMS_CONTENT_RULES` block to `buildCodingWorkspacePrompt` — the single assembler shared by the decopilot engine, claude-code CLI, and codex CLI harnesses — stating:

- CMS content lives in `.deco/blocks/<encoded-key>.json`; edit **only** those files
- **never** edit `blocks.gen.json` / `blocks.gen.ts` / `meta.gen.json` / any `*.gen.*` (derived, regenerated by the dev watcher and every production build)
- no other file needs touching: the `decoVitePlugin` watcher handles preview HMR (delta → `blocks.gen.json` + POST `/.decofile` + full-reload), and production reads `.deco/blocks/` directly
- after a content edit, seed these rules into the repo's AGENTS.md **once** (create if missing, skip if already documented) so non-studio tooling follows the same contract

The block is self-gating text ("when the repository contains a `.deco/` directory…") because the prompt is assembled server-side with no sandbox fs access — inert for non-deco workspaces and prompt-cache-stable.

## Notes for reviewers

- Studio already treats `blocks.gen.json` as generated noise (`stripGeneratedFilesFromDiff`), consistent with this rule.
- The AGENTS.md seeding means a one-time extra `AGENTS.md` diff in customer repos on their next content edit. `isDecoOnlyDiff` publish gating may want to whitelist `AGENTS.md` — flagged for discussion, not changed here.

## Testing

- `bun test packages/harness/src/coding-workspace-prompt.test.ts` — 8 pass, including a new test asserting the rules render for all workspace kinds.

---
## Summary by cubic
Teach agents the correct Deco CMS content files to edit and avoid silent data loss. Edits target `.deco/blocks/*.json` only; generated artifacts like `blocks.gen.json` are off-limits. Rules are now appended only when the repo has a `.deco/` directory.

- **New Features**
  - Gate `DECO_CMS_CONTENT_RULES` on a real `.deco/` check via a new `isDecoSite` flag in `buildCodingWorkspacePrompt` (removed self-gating prose).
  - CLI harnesses (`claude-code`, `codex`): set `isDecoSite` using `localWorkspaceIsDecoSite(cwd)` (sync check for `/repo`).
  - Decopilot engine: set `isDecoSite` using `sandboxIsDecoSite` (cluster `test -d .deco` via sandbox-fs), gated on `user + branch`, and fail open on errors.
  - Tests: added `coding-workspace-deco.test.ts`; expanded `coding-workspace-prompt.test.ts` to assert inclusion only when `isDecoSite` is true.
  - Rules also seed a short note in `AGENTS.md` once if missing.

<sup>Written for commit 69e6eb5c6bbb6038522d7dcea7b1ecb5873d7a39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4485?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/harnesses/decopilot/build-agent-system-prompt.ts
+++ b/apps/mesh/src/harnesses/decopilot/build-agent-system-prompt.ts
@@ -24,6 +24,7 @@ import {
   type CodingWorkspacePromptInput,
 } from "@decocms/harness/coding-workspace-prompt";
 import { buildOrgFilesystemPrompt } from "@/api/routes/decopilot/constants";
+import { sandboxIsDecoSite } from "./built-in-tools/cluster-sandbox-fs";
 import type { GithubRepo } from "@decocms/mesh-sdk";
 import type { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import {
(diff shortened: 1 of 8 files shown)


pedrofrxncx (changes requested):
pls go further on this; only append this prompt if the agent is attached to a sandbox that is running a deco site

pedrofrxncx (approved):

PR #4488, opened 2026-07-13
feat(agents): link an agent's system prompt to an org-fs file

viktormarinho:
Lets an agent's system prompt be sourced from a linked markdown/txt file in the org filesystem (`metadata.instructionsFile`) instead of only the inline `metadata.instructions` mirror. The runtime reads the linked file at agent-load time (same async seam as the skills catalog), so both the cluster engine and the sandbox/desktop daemon pick it up. Editing the prompt in the UI writes through to the file (guarded so a stale-mirror autosave never clobbers an external file edit), and files on readonly `public-*` (GitHub-synced) volumes lock the editor.

## Screenshots/Demonstration
No UI screenshots captured this session — the agent settings "Instructions" section gained a "Link file" control next to Improve/template actions; linking shows a chip with the file name (and a lock icon when the volume is readonly).

## How to Test
1. Open an agent's settings, click "Link file" under Instructions, and pick (or search for) a markdown/text file from the org filesystem.
2. Edit the prompt in the textarea and confirm the linked file's content updates (check via the Library or the fs API).
3. Edit the file externally (e.g. through a GitHub-synced `public-*` set) and reload the agent — confirm the new file content is served as the system prompt and the editor is locked read-only for public volumes.

## Migration Notes
No database migration — `instructionsFile` is a new optional key on the existing JSON `metadata` column.

## Review Checklist
- [x] No breaking changes

---
## Summary by cubic
Link an agent’s system prompt to a markdown/text file in the org filesystem via `metadata.instructionsFile`. The runtime reads the file at load time, edits write back to it, and public (GitHub‑synced) volumes are read‑only.

- **New Features**
  - Runtime prefers the linked file over `metadata.instructions` and reads it at agent load in both the cluster and sandbox.
  - UI adds “Link file” for Instructions; shows a chip when linked, and locks the editor for files on `public-*` volumes; unlink keeps the current text inline.
  - Edits write through to the linked file with safeguards to avoid clobbering external edits; prompt file content is refetched after save.
  - SDK/types updated to include `metadata.instructionsFile`; added E2E tests covering runtime source, write‑through, and external edit precedence.

- **Migration**
  - No DB changes. Adds optional `metadata.instructionsFile` in the existing JSON metadata.

<sup>Written for commit 09e0e6bcebea58b3cc8c69313aaedd76de803c8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4488?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/mcp-clients/virtual-mcp/instructions-file.ts
+++ b/apps/mesh/src/mcp-clients/virtual-mcp/instructions-file.ts
@@ -1,0 +1,39 @@
+/**
+ * Resolves the agent's system prompt from its linked org-fs file
+ * (`metadata.instructionsFile`). Read here (async, factory-side) — NOT inside
+ * the synchronous `getInstructions()` — and stashed on the client options,
+ * the same seam as the skills block, so it reaches both the cluster engine
+ * and the sandbox/desktop daemon. Best-effort: any failure degrades to the
+ * inline `metadata.instructions` mirror, never a failed client creation.
+ */
+
+import type { StudioContext } from "../../core/studio-context";
+import {
+  buildPublicOrgFs,
+  isPublicVolume,
+} from "../../file-storage/public-sets";
+import type { VirtualMCPEntity } from "../../tools/virtual/schema";
+
+export async function loadInstructionsFileText(
+  ctx: StudioContext,
+  virtualMcp: VirtualMCPEntity,
+): Promise<string | null> {
+  const ref = virtualMcp.metadata?.instructionsFile;
+  if (!ref) return null;
+  try {
+    // `public-*` volumes live under the shared public scope, not the org's.
+    const orgFs = isPublicVolume(ref.volume)
+      ? buildPublicOrgFs(ctx)
+      : ctx.orgFs;
+    if (!orgFs) return null;
+    return new TextDecoder().decode(await orgFs.read(ref.volume, ref.path));
+  } catch (err) {
+    console.warn("[virtual-mcp] failed to read linked instructions file", {
+      virtualMcpId: virtualMcp.id,
+      volume: ref.volume,
+      path: ref.path,
+      err,
+    });
+    return null;
+  }
+}
--- a/apps/mesh/src/tools/virtual/update.ts
+++ b/apps/mesh/src/tools/virtual/update.ts
@@ -12,6 +12,7 @@ import {
   requireOrganization,
 } from "../../core/studio-context";
 import { writeAgentPrompts } from "../../file-storage/agent-prompts";
+import { isPublicVolume } from "../../file-storage/public-sets";
 import { VirtualMCPEntitySchema, VirtualMCPUpdateDataSchema } from "./schema";
 import { requireOrgAdminForPinnedField } from "./require-org-admin-for-pin";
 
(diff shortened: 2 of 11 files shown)


pedrofrxncx (commented):
Reviewed this pretty carefully — traced the full data flow and I like the design. Confirmed the override reaches **both** runtimes (cluster via `tools.ts` → `getInstructions()` → `buildAgentSystemPrompt`, and the sandbox/endpoint via `virtual-mcp.ts`), and the write-through gate (`incoming ≠ stored mirror` AND `≠ file`) genuinely prevents stale-autosave clobber and link-time churn — I hand-traced all four E2E scenarios and they hold. Public-volume readonly is enforced in both the UI and the write-through, and unlink preserving inline text is a nice touch.

Leaving a few notes inline. **None are blockers** — mostly defense-in-depth / minor consistency stuff. The only one I'd love a quick decision on is the fs-write permission note, since it's latent rather than exploitable today. Great PR overall. 🙌

apps/mesh/src/tools/virtual/update.ts
```diff
@@ -79,6 +80,37 @@ export const COLLECTION_VIRTUAL_MCP_UPDATE = defineTool({
...
+      incomingInstructions !== existing.metadata?.instructions &&
+      !isPublicVolume(fileRef.volume) &&
+      ctx.orgFs
+    ) {
+      const fileText = await ctx.orgFs
+        .read(fileRef.volume, fileRef.path)
+        .then((bytes) => new TextDecoder().decode(bytes))
+        .catch(() => null);
+      if (fileText !== incomingInstructions) {
+        await ctx.orgFs.write(
```
pedrofrxncx, line 105: Two small notes on the write-through, neither blocking:

**1. Permission scope (defense-in-depth).** This writes via `ctx.orgFs.write(...)` after only `ctx.access.check()` (the agent-update permission), whereas the HTTP fs PUT route gates writes on `ORG_FS_WRITE`. Since `fileRef` is fully caller-supplied, a user who can edit an agent but lacks fs-write could technically overwrite any file in the org by pointing `instructionsFile` at it. Today this is harmless because `ORG_FS_WRITE` is basic-usage (every member has it — see the ACL note at the top of `org-fs.ts`), so there's no live exploit. But that file explicitly anticipates per-volume ACLs, and this becomes a real bypass the moment they land. `writeAgentPrompts` is the precedent, but it writes a fixed agent-scoped path — this one's arbitrary. Might be worth a TODO here, or routing through the same `ORG_FS_WRITE` check.

**2. No `notifyOrgFsChange`.** The HTTP PUT route fires it so other Library viewers / the recent-files feed refresh live. This write skips it, so a prompt edit landing on the file won't propagate to other clients (the editing client self-invalidates its own query, but that's it). Minor.

apps/mesh/src/mcp-clients/virtual-mcp/instructions-file.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!ref) return null;
+  try {
+    // `public-*` volumes live under the shared public scope, not the org's.
+    const orgFs = isPublicVolume(ref.volume)
+      ? buildPublicOrgFs(ctx)
+      : ctx.orgFs;
+    if (!orgFs) return null;
+    return new TextDecoder().decode(await orgFs.read(ref.volume, ref.path));
+  } catch (err) {
+    console.warn("[virtual-mcp] failed to read linked instructions file", {
```
pedrofrxncx, line 31: Not blocking — two minor observations:

- This runs a fresh org-fs read (S3 GET) on every client creation for file-linked agents, uncached. That's the correct tradeoff since "external edit wins" needs freshness (and it mirrors `skillsBlock`), just flagging the per-run latency/cost.
- The `console.warn` logs the full `err` object every run when a linked file is missing/unreadable — a deleted linked file will spam this once per run. Might drop to `debug` or trim the payload.

apps/mesh/src/web/views/virtual-mcp/index.tsx
```diff
@@ -1153,6 +1156,29 @@ function VirtualMcpDetailViewWithData({
...
+  const promptFileReadOnly =
+    promptFile !== null && publicSetOf(promptFile.volume) !== null;
+  const { data: promptFileText } = useOrgFsReadText(
+    promptFile?.volume ?? null,
+    promptFile?.path ?? "",
+  );
+  const [promptDraft, setPromptDraft] = useState<string | null>(null);
+  // Post-save echo: the refetched file matches the draft — drop it so later
+  // external file updates show through. Render-phase adjust, not an effect.
+  if (promptDraft !== null && promptFileText === promptDraft) {
```
pedrofrxncx, line 1174: Nit, not blocking: the draft-clear leans on exact string equality (`promptFileText === promptDraft`). It's safe today because the write and read round-trip verbatim (no normalization on either side), but if any normalization ever creeps into the fs write path, the draft would never clear and later external edits wouldn't surface in the editor. A one-line comment naming that dependency would save a future debugging session.

PR #4573, opened 2026-07-15
fix(dispatch-queue): cap per-worker DBOS concurrency on heavy queues

pedrofrxncx:
## Summary

The hosted-harness and background-tools DBOS queues set only **per-partition** concurrency (`1` per thread, `5` per org). Nothing bounded how many workflows a **single worker pod** dequeues across all active partitions — so one pod could run an unbounded number of concurrent agent loops.

Measured impact on prod (`deco-studio`, 7d): worker working-set is ~300MB median but the tail reaches **677 / 921 / 1794 MB** — the 1.8GB pod sat right under the 2Gi limit and OOM-restarted. It also defeats the new KEDA queue-depth autoscaler: backlog piles onto existing pods instead of spilling to new ones.

## Change

Add `workerConcurrency` (a per-process cap, orthogonal to the per-partition cap) to the two memory-heavy queues:
- `decopilot-hosted-harness` → **3** (the in-process agent loop: LLM streaming + tool calls)
- `background-tools` → **4** (generate_image et al.)

thread-gate and automations are left uncapped — they're light dispatch/gate workflows that spawn children and await.

At `workerConcurrency=3` a loaded hosted-harness pod stays near ~1GB, and bursts beyond that trigger KEDA to add pods.

## Testing
- `tsc --noEmit` passes
- `biome format` clean
- No runtime behavior change under low concurrency; the cap only engages when a single pod would otherwise run >3 hosted runs at once.

---
## Summary by cubic
Cap per-worker DBOS concurrency on heavy queues to prevent pod OOMs and let bursts trigger KEDA queue-depth scale-out. This bounds concurrent agent loops per pod while keeping per-partition fairness.

- **Bug Fixes**
  - Add `workerConcurrency` for `HOSTED_HARNESS_QUEUE` and `BACKGROUND_TOOLS_QUEUE`.
  - Set `HOSTED_HARNESS_WORKER_CONCURRENCY = 3` and `BACKGROUND_TOOLS_WORKER_CONCURRENCY = 4`; export and wire into registration.
  - Keep existing per-partition caps (`1` per thread, `5` per org); thread-gate and automations remain uncapped.
  - Stabilizes per-pod memory (~1 GB under load) and spreads backlog to new pods; no change at low concurrency.
  - Re-baseline DBOS workflow source snapshot; only top-level constants added, no workflow-body changes, recovery-compatible (no version bump).

<sup>Written for commit f56eccd957fca4bdc1ca6a031714527fcd8d3a9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4573?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/api/app.ts
+++ b/apps/mesh/src/api/app.ts
@@ -163,6 +163,7 @@ import {
 import {
   HOSTED_HARNESS_PARTITION_CONCURRENCY,
   HOSTED_HARNESS_QUEUE,
+  HOSTED_HARNESS_WORKER_CONCURRENCY,
   setHostedHarnessRuntime,
   setThreadGateRuntime,
   THREAD_GATE_PARTITION_CONCURRENCY,
@@ -795,6 +796,7 @@ import { type SSEEvent, sseHub } from "../event-bus";
 import {
   BACKGROUND_TOOLS_PARTITION_CONCURRENCY,
   BACKGROUND_TOOLS_QUEUE,
+  BACKGROUND_TOOLS_WORKER_CONCURRENCY,
   setBackgroundToolRuntime,
 } from "@/harnesses/decopilot/background-tool-workflow";
 import { abortBackgroundJobs } from "@/harnesses/decopilot/background-abort-registry";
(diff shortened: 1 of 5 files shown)


pedrofrxncx:
Closing — the approach doesn't work.

I intended `workerConcurrency` to cap how many concurrent runs a single worker **pod** pulls across all threads (to bound per-pod memory). But reading the SDK dispatch (`findAndMarkStartableWorkflows` in dbos-sdk 4.21.6), it's called **per partition key**, and both `concurrency` and `workerConcurrency` are enforced **within a partition**: `maxTasks = workerConcurrency - countRunning(queue, partitionKey)`.

So for a partitioned queue `workerConcurrency` is a *per-partition* limit — same scope as `concurrency` — not a per-process global cap. On the `concurrency=1` queues (thread-gate, hosted-harness) it's also rejected outright by `validateQueueParams` (`workerConcurrency must be <= concurrency`), which is exactly the boot crash CI caught:

```
error: concurrency must be greater than or equal to workerConcurrency
  at registerQueue (dbos-sdk/dist/src/dbos.js:1783)
```

There is no per-executor global workflow cap in this SDK version that composes with per-partition serialization. The existing per-partition limits (1/thread, 10/org, 5/org) are sensible for the model; per-pod memory is bounded by the 2Gi limit + KEDA (CPU + queue-depth) replica scaling instead. Nothing to change here.

PR #4702, opened 2026-07-17
fix(chat): tolerate orphaned tool output in stream reconstruction

pedrofrxncx:
## Summary

Reported on staging: a chat bricks with `No tool invocation found for tool call ID "toolu_…"`. The error is thrown in the browser by the AI SDK's `readUIMessageStream` when a `tool-output-available` / `tool-output-error` / `tool-input-delta` chunk references a tool call whose `tool-input-*` part is neither in the seed message nor earlier in the stream.

That invariant breaks whenever a run is **reconstructed after abnormal termination** — the tool's output survives but its input does not:

- a **user-desktop (link) run** whose desktop/dev-server drops mid-tool-call (the reported case: `LINK_CURRENT_GET`, `503 Dev server not running`, `502 "Is the computer able to access the url?"`, then a `Ghost run detected, force-failing`),
- a cluster-hosted **ghost-run force-fail** (pod terminated mid-step),
- a JetStream **subject purge** in the reconnect window, or a **retention gap**.

The existing repro (`thread-connection-replay.test.ts` → *"refresh AT completion … throws"*) documents this exact failure.

## Fix

Add `tool-invariant-guard.ts`: a small, transport-agnostic guard placed between the chunk stream and `readUIMessageStream` in `foldSubStream`. It tracks which tool call ids already have an invocation (seeded from the folded message's existing tool parts, then updated as `tool-input-*` chunks pass through) and, before any orphaned output/delta, injects a synthetic `tool-input-available` (placeholder name `unknown`, empty input). The tool then renders as a completed call with an unknown name instead of aborting the whole run.

It's a pass-through for healthy streams (an id seen via a real `tool-input-*` or present in the seed is never synthesized), so it only changes behavior on the degraded path.

Because the throw is in the browser and both cluster-hosted and desktop-link runs fold through the same client reader, this one guard covers every trigger.

## Testing

- Inverted the documented failure: the same orphan post-purge chunks that `readUIMessageStream` rejects now fold cleanly through the guard (tool part present, `output-available`, output preserved).
- 6 pure-unit tests for the guard (pass-through pair, synthesize-before-orphan, no-synthesize-when-seeded, synthesize-once-per-id, non-tool chunks untouched, `toolCallIdsInMessage`).
- 13/13 pass; `tsc --noEmit` clean; lint 0 errors; `bun run fmt` applied.

## Out of scope

The underlying triggers (staging pods cycling on a 115s drain-timeout force-exit; desktop links dropping) are separate infra issues. This PR makes the chat *survive* them rather than fixing why runs get orphaned.

---
## Summary by cubic
Prevent chat and the subtask panel from crashing or truncating when reconstructing runs that lost tool-input chunks. We synthesize a minimal tool-input before orphaned tool output so the UI keeps working instead of throwing “No tool invocation found for tool call ID …”.

- **Bug Fixes**
  - Added a transport-agnostic guard (`tool-invariant-guard.ts`) that tracks seen tool call IDs and injects a synthetic `tool-input-available` before any orphaned `tool-output-*` or `tool-input-delta`.
  - Wrapped the main thread reader (`ThreadConnection`) and the subtask live tail (`use-subtask-stream`) in `guardToolInvariant` so both the chat and subtasks are protected.
  - Pass-through for healthy streams; only activates on reconstructed/abnormally terminated runs (ghost-run force-fail, desktop link drop, JetStream purge, retention gap).
  - Added focused unit tests and a replay test confirming the guard preserves tool output and avoids reader exceptions.

<sup>Written for commit bdcbfc94058414b69b03d4771b919a2abc2503c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/4702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/src/web/components/chat/message/parts/tool-call-part/use-subtask-stream.ts
+++ b/apps/mesh/src/web/components/chat/message/parts/tool-call-part/use-subtask-stream.ts
@@ -1,5 +1,6 @@
 import { useEffect, useState } from "react";
 import { readUIMessageStream, type UIMessage, type UIMessageChunk } from "ai";
+import { guardToolInvariant } from "../../../store/tool-invariant-guard";
 
 /** Parse an SSE byte stream into the `UIMessageChunk`s the AI SDK reader folds. */
 function sseToChunkStream(
@@ -74,8 +75,11 @@ export function useSubtaskStream(args: {
           },
         );
         if (!resp.ok || !resp.body) return;
+        // Same tool-invocation guard as the main thread reader: a reconstructed
+        // subtask stream can deliver a tool's output without its input part,
+        // which would otherwise throw and silently truncate the subtask view.
         for await (const msg of readUIMessageStream({
-          stream: sseToChunkStream(resp.body),
+          stream: guardToolInvariant(sseToChunkStream(resp.body), undefined),
         })) {
           if (abort.signal.aborted) return;
           setMessages([msg]);
--- a/apps/mesh/src/web/components/chat/store/thread-connection-replay.test.ts
+++ b/apps/mesh/src/web/components/chat/store/thread-connection-replay.test.ts
@@ -29,6 +29,7 @@
 import { describe, expect, test } from "bun:test";
 import { readUIMessageStream } from "ai";
 import type { UIMessage, UIMessageChunk } from "ai";
+import { guardToolInvariant } from "./tool-invariant-guard";
 
 const TOOL_CALL_ID = "tc_replay_test";
 const MESSAGE_ID = "msg_replay_test";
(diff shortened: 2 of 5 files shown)


pedrofrxncx:
## Reviewed — good to merge ✅

Read the guard, the wiring, and both tests. The design holds up:

**Correctness**
- The seed feeding `readUIMessageStream({ message: seed })` is the *same* object feeding `guardToolInvariant(sub, seed)`, so the guard's known-id set is exactly the reader's set of pre-existing invocations — no false synthesis on healthy/seeded streams.
- `CREATES_INVOCATION` (`tool-input-start`/`-available`) vs `REQUIRES_INVOCATION` (`tool-input-delta`, `tool-output-available`, `tool-output-error`) matches exactly the three chunk types `readUIMessageStream` throws `No tool invocation found` on. Synthesize-once-per-id is enforced by adding to `known` on the synthetic path.
- Genuinely transport-agnostic and a pure pass-through for healthy streams — behavior only changes on the degraded/reconstructed path, so blast radius is minimal.

**Tests** — the documented failure (`refresh AT completion … throws`) is *inverted* into a passing case (orphan output now folds to a `tool-unknown` part carrying the recovered output), plus 6 focused unit tests. 13/13 pass, `tsc` clean, lint 0 errors.

**One improvement pushed** (`bdcbfc9`): the PR body claims *one guard covers every trigger*, but there was a **second** `readUIMessageStream` consumer — `use-subtask-stream.ts` (the subtask panel's live tail) — that folds the same reconstructable streams and was unguarded. It wouldn't crash there (the surrounding `catch {}` swallows the throw) but it would **silently truncate** the subtask view at the orphan point, dropping the tool output and everything after. Routed it through the same `guardToolInvariant` so the claim actually holds across every reader.

Enabling auto-merge.

PR #5117, opened 2026-07-23
feat(billing): per-seat foundation — org billing tables + seat gate (dormant)

viktormarinho:
## Summary

Phase-2 opener of the per-seat monetization plan: data model + enforcement seam, **fully dormant** — behind `STUDIO_BILLING_ENFORCED` (default off, declared off in [deco-apps-cd#175](https://github.com/decocms/deco-apps-cd/pull/175)), with migration 139 backfilling every existing org `legacy = true`. Deploying this changes behavior for no one.

- **migration 139**: `organization_billing` (billing identity — platform-written only, deliberately NOT in org `metadata`, which org admins can write via `ORGANIZATION_UPDATE`; billing there would let any admin flag their own org legacy) + `organization_paid_seat` (presence = paid seat; absence = free). New orgs get `legacy = false` from the org-creation hook; a missing row fails **open** (treated legacy) so a hook failure can't brick an org.
- **Seats ≠ roles**: seats are monetization, orthogonal to Better Auth governance. An org **owner** can hold a free seat (the report-funnel onboarding user), so the seat gate fires in `AccessControl.check()` **before** the admin/owner bypass.
- **Free seat = sees everything, changes nothing**: tools declaring `annotations.readOnlyHint: true` pass (fail-closed — an unannotated read tool gets annotated, not the gate widened); route-level resource keys have a read allowlist (`ORG_FS_READ`, not `ORG_FS_WRITE`).
- **`PaidSeatRequiredError`** extends `ForbiddenError` (every existing handler keeps returning 403 untouched) and carries the **`[PAID_SEAT_REQUIRED]` message prefix as the wire contract** — same convention as `[CREDITS]`/`is-credit-error.ts`, so the future paywall UI distinguishes it from a plain permission 403 across every transport (REST, MCP tool error). CTA branching (subscribe vs "ask your admin") uses the role the frontend already has.
- `resolveOrgFromPath` loads billing + seat state (two PK lookups, members only, flag on only) and arms the gate for the request.

## Testing

- 6 new unit tests in `access-control.test.ts` (32 pass): owner-with-free-seat blocked, `readOnlyHint` passes, basic-usage tools gated, `ORG_FS_READ` passes / `ORG_FS_WRITE` blocked, gate-off default fully open, `ForbiddenError` inheritance + prefix contract.
- Full `bun test` green; `bun run check` (all workspaces) green; `bun run lint` no new warnings; `knip` clean; `bun run fmt` applied.
[...]

PR #5128, opened 2026-07-23
feat(billing): seat management tools + change-log (invoiced orgs)

viktormarinho:
## Summary

Task 2.5 of the per-seat plan (builds on #5117). Backend for seat management — the piece that unlocks managing contract (`invoiced`) orgs in-product; UI (2.6) and gateway benefits sync (2.8) come next.

- **Migration 140** `seat_change_log`: append-only seat transitions, written in the same transaction as the `organization_paid_seat` change. For `invoiced` orgs this IS the billing source (end-of-cycle boleto reads who held a paid seat when), so rows are appended **only for actual transitions** — a no-op "set paid on already-paid" writes nothing.
- **`OrganizationBillingStorage`** (`ctx.storage.organizationBilling`): `getBilling`, `listPaidSeatUserIds`, `setSeats` (transactional; validates every target is a current member — any unknown user rejects the whole batch, so a stale UI can't bill a ghost), `releaseSeatOnMemberRemoval`.
- **`ORGANIZATION_SEATS_GET`** — `readOnlyHint: true`, so free-seat members can render the members/billing page through the seat gate. **`ORGANIZATION_SEATS_SET`** — `members:manage` capability (seats change who the org pays for; same trust tier as removing the member). SET gates on `billing_mode`: `invoiced` applies directly; `self_serve` errors until the Stripe checkout path (phase 3); `legacy` errors (no seats).
- **Lifecycle pairing**: `ORGANIZATION_MEMBER_REMOVE` now releases the removed member's paid seat + logs it (fail-soft — release failure never fails the completed removal). `boundAuth.removeMember` now returns Better Auth's result (it carries the removed `userId`).

## Testing

- New real-Postgres integration suite (`organization-billing.integration.test.ts`, `Storage Integration` workflow): transition-exact logging, replay idempotency (nothing applied/logged), batch atomicity on unknown members, seat release on removal, `getBilling` mapping + missing-row null. 4/4 green locally against a migrated Postgres 16.
- Registry/core suites green (89), full-workspace `bun run check` green, `knip` clean, `bun run fmt` applied. The 8 pre-existing env-dependent tool-test failures are identical on main (verified via stash).
- Still dormant end to end: without `STUDIO_BILLING_ENFORCED` nothing consults seats, and no org has `billing_mode = invoiced` until flipped manually for the pilot.

---
## Summary by cubic
Adds backend seat management for invoiced orgs with transactional seat changes and an append-only change log. Seat lifecycle now runs at Better Auth membership hooks to auto-release/reset seats and make orphan rows harmless; enforcement stays behind `STUDIO_BILLING_ENFORCED`; UI and self-serve checkout come later.

- New Features
  - `OrganizationBillingStorage`: `getBilling`, `listPaidSeatUserIds` (JOINs `member` so stale rows never count), `setSeats` (atomic, member-validated; logs only real transitions), `releaseSeat`.
  - Tools: `ORGANIZATION_SEATS_GET` (readOnlyHint) and `ORGANIZATION_SEATS_SET` (under `members:manage`, applies only to `invoiced`; `self_serve` errors, `legacy` has no seats).
  - Membership hooks: `afterRemoveMember` releases any paid seat; `afterAddMember` drops stale rows on rejoin. Both fail-soft.

- Migration
  - New table: `seat_change_log` (append-only) + index on `organization_id, created_at`.
  - Registered as migration `140`; no behavior change until the flag is enabled.

<sup>Written for commit fa0eb12baf953f0b2e0f8d965c39637c6b7473c6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/mesh/migrations/140-seat-change-log.ts
+++ b/apps/mesh/migrations/140-seat-change-log.ts
@@ -1,0 +1,36 @@
+import { type Kysely, sql } from "kysely";
+
+/**
+ * Append-only log of seat transitions (paid <-> free), written in the same
+ * transaction as the organization_paid_seat change. For `invoiced`
+ * (contract) orgs this IS the billing source: end-of-cycle invoicing reads
+ * who held a paid seat when, so rows are only appended for ACTUAL
+ * transitions (a no-op "set paid on already-paid" writes nothing).
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createTable("seat_change_log")
+    .addColumn("id", "uuid", (col) =>
+      col.primaryKey().defaultTo(sql`gen_random_uuid()`),
+    )
+    .addColumn("organization_id", "text", (col) =>
+      col.notNull().references("organization.id").onDelete("cascade"),
+    )
+    .addColumn("user_id", "text", (col) => col.notNull())
+    .addColumn("seat", "text", (col) => col.notNull())
+    .addColumn("changed_by", "text", (col) => col.notNull())
+    .addColumn("created_at", "timestamptz", (col) =>
+      col.notNull().defaultTo(sql`now()`),
+    )
+    .execute();
+
+  await db.schema
+    .createIndex("idx_seat_change_log_org_created")
+    .on("seat_change_log")
+    .columns(["organization_id", "created_at"])
+    .execute();
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropTable("seat_change_log").execute();
+}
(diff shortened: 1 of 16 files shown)


viktormarinho:
Design hardening after review discussion: **what happens if the server dies between the Better Auth removal and the seat release?**

Durable execution (DBOS) is the wrong tool for this one — the removal commits inside Better Auth's tables, so a workflow enqueue can't be atomic with it; a crash before `startWorkflow` leaves the same orphan. Instead the orphan is now **harmless by construction**:

1. Every paid-seat count/list **JOINs `member`** — a stale row (member removed, release never ran) can never count toward billing or the gateway allowance.
2. Seat release moved from the tool into the **`afterRemoveMember` organization hook** — the canonical boundary, covering every removal path (tool, direct BA endpoint, etc).
3. New **`afterAddMember` hook** drops any stale paid row on (re)join — invites always land free, so a crashed release can't resurrect a paid seat when the user comes back.

Integration test now simulates the crash window explicitly (orphan row present → excluded from every count). The `removeMember` return-type change is reverted (no longer needed).

PR #5215, opened 2026-07-24
feat(cms): fast preview via preview server + branch-aware new chat

tlgimenes:
## Summary

CMS **Fast Preview** (render drafts against an always-on preview server, no dev-server wait), a reworked agent-settings **CMS** section, a **New chat** fix (always creates, on the current branch), and a set of **header UX** improvements (branch label, truncation priority, and small-screen icon-only actions).

---

### 1. Fast Preview
- Renders the current working-tree draft against an always-on **preview server** via `/live/previews` (decofile pushed as a per-request, ALS-isolated override) — no waiting for the sandbox dev server.
- `buildInstantPagePreviewUrl()` builds the cross-origin GET URL with a URL-size cap; past it, falls back to the published route.
- Fast Preview **never** swaps to the sandbox surface — it previews *only* on the preview server for the whole lifecycle (even once the dev server is up / crashed). The "waking" pill tracks the sandbox **itself** (FS/handle → `blocks.gen.json` readable), not the dev server.
- Fast Preview **off**: prior behavior kept — published site while booting, then swap to the sandbox once its dev server runs.
- URL-bar domain follows what's in the iframe (`buildPreviewLabel`); **open-in-new-tab** is hidden unless a real sandbox iframe is shown.

### 2. Settings — CMS section
- New **CMS** section: **Preview** (preview-server URL + Fast Preview switch, gated on the URL) and **Publishing** (moved in; extracted `PublishPolicyField`). Business-friendly copy, en + pt-br.

### 3. New chat — always creates, on the current branch
- The **New chat** button now **always creates** a fresh thread (removed the empty-chat reuse short-circuit from both entry points), on the **agent + branch currently being viewed** (was hardcoded `staging`). `createNewTask` threads the branch through (panel-actions + keyboard paths). "Open agent X" flows still reuse an empty chat, branch-agnostically. No backend change — the wire schema already accepts `branch`.

### 4. Header / small-screen UX
- **Branch button**: always shows the current branch (truncated); it now sits in its own shrinkable slot so it **truncates before the address bar** (which is `flex-1`, shrinks last), while the publish/⋯ actions stay `shrink-0` and never clip.
- Below **`lg` (< 1024px)** the header collapses to **icon-only + tooltip**: the branch button, the primary action button (icon keyed off `button.action` — `GitPullRequest` / `RefreshCw01` / `AlertTriangle` / `CheckCircle` / `MessageCircle01`), the side **Publish** button (`Upload01`), and the **Develop/Live** toggle (`Code01` / `Globe01`). Loading pills collapse to spinner-only; plain status pills (Up to date / Published / Awaiting review) keep their text. Desktop is unchanged.

## Testing
- Unit: `preview-display.test.ts` (13), `preview-label.test.ts` (7), `section-preview-url.test.ts`, `reusable-new-chat.test.ts` (8).
- `bun run check` (web + shared) and `bun run lint` pass; formatted with Biome. Layout/copy-only changes have no unit tier.

## Notes / follow-ups
- In Fast Preview mode the sandbox iframe is never shown, so same-origin **in-iframe editor overlays** (visual editor, click-to-select) don't activate on the cross-origin preview server; **Blocks-panel** editing (via `blocks.gen.json`) still works.
- Instant render is **static** (`renderToString`) — no hydration/JS interactivity/in-page navigation. Lifting the URL-size cap (POST proxy / draft cookie) is a natural follow-up for large sites.
- `merge-split` ("Publish to `<base>`") renders via `MergeSplitButton` and is still text below `lg` (not yet iconified).
- Possible follow-up (discussed, not in this PR): when Fast Preview is on, defer `bun install` + the dev script at boot (the decofile is a dep-free FS merge) and boot them lazily on first need — behind a default-off flag.

### DIFF
--- a/apps/api/src/link-daemon/index.ts
+++ b/apps/api/src/link-daemon/index.ts
@@ -186,6 +186,9 @@ export async function startLinkDaemon(
               ? { path: config.workload.packageManagerPath }
               : {}),
           };
+          if (config.workload.productionUrl) {
+            application.productionUrl = config.workload.productionUrl;
+          }
         }
         const payload: Record<string, unknown> = { application };
         const operator = normalizeCoAuthorIdentity(config.operator ?? null);
--- a/apps/api/src/link-daemon/user-desktop-provider.ts
+++ b/apps/api/src/link-daemon/user-desktop-provider.ts
@@ -89,6 +89,12 @@ export interface Workload {
   devPort?: number;
   /** Subdirectory inside the repo where the package manifest lives. */
   packageManagerPath?: string;
+  /**
+   * Live production URL of the linked site. Forwarded to the daemon so the
+   * `/_deco/fast-preview` route can render the working-tree decofile against
+   * production without the dev server.
+   */
+  productionUrl?: string;
 }
 
 export interface EnsureSandboxInput {
--- a/apps/api/src/tools/sandbox/start.ts
+++ b/apps/api/src/tools/sandbox/start.ts
@@ -592,6 +592,14 @@ async function provisionSandboxInner(
   // runners free to assign a unique dynamic port (user-desktop needs this;
   // multiple sandboxes on the user's machine share the host network and
   // can't all bind 3000).
+  // Live production URL (already normalized to an https URL by the settings
+  // form). Forwarded so the daemon's Fast Preview route can render the draft
+  // against production. Harmless when Fast Preview is off — the daemon only
+  // reads it when that route is hit.
+  const productionUrl =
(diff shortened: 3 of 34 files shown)


tlgimenes:
Split into three focused PRs (same commits, cleanly separated by concern — the file sets are disjoint):

- **Header / branch UI** → #5230
- **New chat: always creates + inherits the current branch** → #5231
- **Fast Preview (daemon-proxy render)** → #5232

Closing this in favor of those.

PR #5310, opened 2026-07-27
fix(sandbox): drop NodeSource — copy Node from the official image instead

viktormarinho:
## Summary

`deb.nodesource.com` 403s intermittently and outages last hours — today's blocked #5305's `docker-smoke` twice, even with the Dockerfile's 5×10s retry loop. The dependency exists only to get `node` + `npm` onto the sandbox image (user repos need them), and there's a strictly more reliable source: the **official `node:26-slim` image**, via multi-stage `COPY`.

- Same supply chain the build already trusts (Docker registry, digest-cached) — one less third-party shell script piped into the build.
- Multi-arch (amd64/arm64) for free, matching the release build.
- `libatomic1` added to the apt list: the official Node binary links it; `oven/bun`'s debian base doesn't ship it (node-slim's does — this was the one integration gap, caught by a local build).
- Corepack/yarn/pnpm logic untouched (Node 26 dist still ships no corepack).
- `NODE_MAJOR` arg still the single place the major is pinned.

## Testing notes

Local build + full toolchain verification inside the image:
`node v26.5.0` · `npm 11.17.0` · `npx` · `corepack 0.35.0` · `yarn` · `pnpm` · `deno 1.46.3` · `bun 1.3.14` — and the same `/health` daemon smoke test docker-smoke runs, green (`bootId` echo verified).

This PR touches `packages/sandbox/**`, so CI's `docker-smoke` builds this exact Dockerfile on this PR — if NodeSource is still down, that's the proof it works (the build no longer contacts it at all).

---
## Summary by cubic
Drops NodeSource from the sandbox image by copying `node`/`npm` from `node:26-slim`, removing flaky 403s and stabilizing CI. Makes timer-driven tests fully deterministic, adds an injectable sleep to `waitForPort`, and bounds the unit test job to 10 minutes to avoid zombie runs.

- **Dependencies**
  - Copy `node` and `npm` from `node:26-slim`; remove NodeSource setup.
  - Add `libatomic1` for the official Node binary.
  - Keep `NODE_MAJOR` as the single pin; install `corepack@0.35.0` via `npm` if missing; enable Corepack.

- **Bug Fixes**
  - Replace real-timer sleeps with manual clocks in heartbeat and SSE tests; assert exact counts and ordering.
  - `wrapStreamWithKeepalive` accepts injectable `KeepaliveTimerFns` (defaults to real timers) and clears the interval on end, cancel, and error.
  - `waitForPort` supports an injectable `sleepFn`; tests are hang-proof (exit if already settled), and the test job is limited to 10 minutes.

<sup>Written for commit ade2f5358fed5e804d151cde5d19ae9c6bee73f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5310?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -114,6 +114,11 @@ jobs:
     needs: changes
     if: needs.changes.outputs.relevant == 'true'
     runs-on: ubuntu-latest
+    # The suite normally finishes in ~1 min. A single hung file under the
+    # parallel runner stalls the whole job silently (per-file output is
+    # buffered, so the culprit prints nothing) — without this bound that's
+    # GitHub's 6h default; it already cost one 50-minute zombie job.
+    timeout-minutes: 10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/apps/api/src/api/routes/decopilot/sse-keepalive.test.ts
+++ b/apps/api/src/api/routes/decopilot/sse-keepalive.test.ts
@@ -1,29 +1,57 @@
-import { describe, it, expect } from "bun:test";
-import { wrapStreamWithKeepalive } from "./sse-keepalive";
+import { describe, expect, it } from "bun:test";
+import {
+  type KeepaliveTimerFns,
+  wrapStreamWithKeepalive,
+} from "./sse-keepalive";
+
+// No real timers anywhere: the interval is a manually-fired ticker and the
+// source is a test-pushed stream. Deterministic under any CI load (the
+// previous small-real-interval style flaked on the loaded parallel runner),
+// and outputs assert exactly.
 
 const enc = new TextEncoder();
 const dec = new TextDecoder();
 
-function makeSource(
-  chunks: Array<{ delayMs: number; bytes: Uint8Array | string } | "close">,
-): ReadableStream<Uint8Array> {
-  return new ReadableStream<Uint8Array>({
-    async start(controller) {
-      for (const item of chunks) {
-        if (item === "close") {
-          controller.close();
-          return;
-        }
-        if (item.delayMs > 0) {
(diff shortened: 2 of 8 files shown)


viktormarinho:
Added a second commit: the `test` job on the first run flaked on `HeartbeatEmitter > emits every intervalMs while armed` — a real-timer test whose 3.8-windows-for-3-emits margin didn't survive the 4-vCPU runner under the parallel unit suite (#5292's CPU contention; the exact failure mode the test's own comment warned about). Widened to 5.8 windows, same ≥3 assertion. `docker-smoke` on the first run was already green — built during the ongoing NodeSource outage, which is the point of this PR.

viktormarinho:
Third commit replaces the margin-widening band-aid with what it should have been: **zero real timers in the three timer-driven test files**. `HeartbeatEmitter` and `withLivenessHeartbeat` already had an injectable `sleepFn` (unused by the old tests); `wrapStreamWithKeepalive` gains an injectable `KeepaliveTimerFns`. Tests now elapse windows / fire ticks manually — no `sleep()` anywhere, CI load can't flake them, and assertions got strictly stronger (exact emit counts, exact byte-stream interleaving, interval-teardown checks that the old tests only hoped for with a 30ms wait). The only waiting left is microtask drains, which don't stretch under CPU contention. Full suite: 6160 pass under `--parallel`.

PR #5328, opened 2026-07-28
fix(web): reload the preview iframe through about:blank on retry

pedrofrxncx:
Follows #5324 (self-heal preview iframe stuck on connection-refused), merged to main just before this run.

## The bug

`useIframeLoadRecovery`'s retry path (`preview-iframe-recovery.ts`) reassigns the iframe's `src` to the exact same URL string it already has to force a reload:

```ts
iframe.src = currentSrc;
```

But in the failure case this watchdog exists to recover from, the iframe never navigated away from that URL in the first place (it's stuck on the browser's connection-refused page while `src` still reads the sandbox URL). Reassigning an unchanged `src` string is a documented no-op in several browsers (Firefox in particular skips the navigation entirely when the value doesn't change; this is the well-known "iframe src reload hack is unreliable" gotcha). So the entire self-heal mechanism silently does nothing in those browsers — it schedules retries forever but the frame never actually reloads.

## The fix

Bounce through `about:blank` before restoring the target URL, so the assignment always differs from the previous value and reliably triggers a real navigation:

```ts
iframe.src = "about:blank";
iframe.src = currentSrc;
```

## Regression test

Added `preview-iframe-recovery.test.ts`, which fakes `setTimeout`/`clearTimeout` with a manually-flushed queue (the real watchdog/backoff delays are 10s/1s) and asserts the iframe's `src` assignment sequence on retry. Confirmed it fails against the pre-fix code (`git stash` + rerun) with the exact same-string no-op it's meant to catch, and passes with the fix.

## To verify

```
bun test apps/web/src/components/sandbox/preview/preview-iframe-recovery.test.ts
```

## Checks run locally

- `bun run fmt`
- `bunx tsc --noEmit` in `apps/web`
- targeted test above (green)

Full CI validates the rest.

---
## Summary by cubic
Fixes the preview iframe self-heal so retries trigger a real reload by bouncing through about:blank. Prevents endless retries on browsers that treat same-URL `src` assignments as no-ops (e.g., Firefox).

- **Bug Fixes**
  - Update `useIframeLoadRecovery` to set `iframe.src = "about:blank"` before restoring the target URL.
  - Add a regression test to assert the `about:blank → target` sequence with fake timers.

<sup>Written for commit a4e7ca1c0aa51b4a76fd856d966b6c2423cf80f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/src/components/sandbox/preview/preview-iframe-recovery.test.ts
+++ b/apps/web/src/components/sandbox/preview/preview-iframe-recovery.test.ts
@@ -1,0 +1,82 @@
+import { setupComponentTest } from "../../../../test/setup";
+setupComponentTest();
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { renderHook, act } from "@testing-library/react";
+import type { RefObject } from "react";
+import { useIframeLoadRecovery } from "./preview-iframe-recovery";
+
+// Fakes setTimeout/clearTimeout with a manually-flushed queue so the
+// watchdog/backoff timers in the hook can be advanced synchronously without
+// waiting out real delays (LOAD_TIMEOUT_MS/RETRY_BASE_MS are 10s/1s).
+function installFakeTimers() {
+  const queue: Array<{ id: number; fn: () => void }> = [];
+  let nextId = 1;
+  const originalSetTimeout = globalThis.setTimeout;
+  const originalClearTimeout = globalThis.clearTimeout;
+  globalThis.setTimeout = ((fn: () => void) => {
+    const id = nextId++;
+    queue.push({ id, fn });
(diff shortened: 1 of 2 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #5340, opened 2026-07-28
feat(sandbox-env): opt-in daemon OTLP metrics endpoint

pedrofrxncx:
Chart side of #5339. That PR gives the daemon a `MeterProvider` that stays dormant without `OTEL_EXPORTER_OTLP_ENDPOINT`; this one lets an operator set it.

```yaml
telemetry:
  otlpEndpoint: ""              # empty = off (default, today's behavior)
  serviceName: studio-sandbox-daemon
```

Renders nothing when empty; injects both env vars when set. Verified with `helm template` both ways, `helm lint` clean.

**Endpoint-agnostic on purpose** — the actual URL belongs in the per-env values in `deco-apps-cd`, not in the chart. That's also why this isn't blocked on deciding *which* collector address to use: the chart change is identical for every option.

## The constraint that makes this non-obvious

`values.yaml` now documents it, because getting it wrong fails **silently** — every export is dropped and the dashboard just stays empty. Sandbox egress is the netinit iptables init container, not a NetworkPolicy, and it appends:

```
lo ACCEPT → conntrack ACCEPT → blockCIDRs REJECT → UDP ACCEPT → TCP ACCEPT → default REJECT
```

The CIDR REJECT fires **before** the port ACCEPTs. A collector on an in-cluster ClusterIP lands in RFC1918, i.e. inside `blockCIDRs`, so it is unreachable **no matter which port** is added to `allowedTCPPorts` — `gateway-otlp` is `172.20.146.131`, inside `172.16.0.0/12`. Reaching a ClusterIP would need a destination-scoped ACCEPT inserted ahead of the `blockCIDRs` loop, which this PR does not add.

An address that already clears the rule needs no netinit change: `gateway-otlp-ingest` is a LoadBalancer on `443:32184/TCP`, and 443 is already in `allowedTCPPorts`. That's the recommended value, pending confirmation of what auth that listener expects.

Also noted in values: enabling this opens a path from pods running untrusted user code to the collector, so a compromised sandbox can emit arbitrary series. The daemon only ever sends fixed instrument names with derived attribute values.

## Notes

- Chart `0.9.33` → `0.9.34`, changelog entry added (the publish workflow skips existing versions).
- README values table updated.
- Merge after #5339 — setting the endpoint against a daemon without the provider is harmless but pointless.

---
## Summary by cubic
Adds an opt-in OTLP metrics endpoint for the sandbox daemon. Default behavior stays the same until you set `telemetry.otlpEndpoint`.

- **New Features**
  - Injects `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_SERVICE_NAME` when `telemetry.otlpEndpoint` is set (chart v0.9.34; README/values updated).

- **Migration**
  - To enable, set `telemetry.otlpEndpoint` to a collector address reachable past netinit. In-cluster ClusterIP is blocked by RFC1918 REJECT; use the collector’s LoadBalancer on 443 or adjust netinit. Enabling allows sandbox pods to reach your collector.

<sup>Written for commit 909f30c46584aaa52087a2226af15894e06cf0e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/deploy/helm/sandbox-env/Chart.yaml
+++ b/deploy/helm/sandbox-env/Chart.yaml
@@ -24,6 +24,11 @@ type: application
 # orgfs-sidecar is SIGKILLed on teardown and its Sandbox CR is already gone) +
 # grants the housekeeper Role `delete` on pods. Additive; existing reap paths
 # unchanged.
+# 0.9.34: opt-in daemon OTLP metrics (telemetry.otlpEndpoint, default
+# empty) — sets OTEL_EXPORTER_OTLP_ENDPOINT / OTEL_SERVICE_NAME so the
+# daemon registers a MeterProvider and pushes. Additive; no change unless
+# set. The endpoint must clear netinit's blockCIDRs REJECT (which fires
+# before the port ACCEPTs) — see values.yaml before picking one.
 # 0.9.4: default studio-sandbox image tag bumped to 1.12.0 (Node.js 26).
 # 0.9.3: opt-in HPA for the warm pool (warmPool.autoscaling.*, default off) —
 # scales SandboxWarmPool via its native scale subresource. No default metric
@@ -47,7 +52,7 @@ type: application
 # 0.8.0: org-fs sidecar (orgFs.*) — privileged rclone mounter + main-container
 # mounts. NOTE: the publish workflow skips existing versions, so this MUST
 # stay ahead of the latest published chart (0.7.5 at time of writing).
-version: 0.9.33
+version: 0.9.34
 # appVersion tracks the studio-sandbox image version (image.tag default).
 appVersion: "1.24.2"
 kubeVersion: ">=1.30.0-0"
--- a/deploy/helm/sandbox-env/README.md
+++ b/deploy/helm/sandbox-env/README.md
@@ -238,6 +238,7 @@ See `values.yaml` for the full set. The most-tuned ones:
 | `netinit.enabled` | `true` | installs the iptables egress policy before user code starts |
 | `disableFsSidecar` | `false` | debug-only opt-out from the mandatory privileged org-fs sidecar |
 | `depsCache.enabled` / `depsCache.golden` | `false` / `false` | opt-in node-local dependency caches |
+| `telemetry.otlpEndpoint` | `""` | opt-in daemon metrics push; endpoint must be reachable past netinit (see `values.yaml`) |
 | `warmPool.enabled` / `warmPool.size` | `false` / `0` | only after measuring cold-start pain |
 | `warmPool.autoscaling.enabled` | `false` | HPA; requires at least one explicit metric |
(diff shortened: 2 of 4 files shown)


pedrofrxncx:
## ⚠️ This chart change cannot reach prod as-is — and it surfaced a live bug

Chasing the grace-period number for the sibling PR turned up chart drift that's worth handling before either merges.

**Prod pins `sandbox-env` `targetRevision: 0.9.5`** (`deco-apps-cd/values.yaml:110`). The tree is at `0.9.33` (`0.9.34` with this PR). So this PR ships a values knob prod won't see until someone bumps `targetRevision` — that bump is a 29-version delta including a default image-tag change, so it wants its own review rather than riding along here.

**The live bug it exposed.** #4561 (`fix(sandbox): make the SIGTERM git-sync survive to completion`, 2026-07-14) raised `terminationGracePeriodSeconds` 30 → 90 in `values.yaml` but **did not bump `Chart.yaml`** — it stayed at `0.9.5`. The publish workflow skips existing versions (the `0.9.2` changelog entry documents this exact trap happening before), so that change was never published. Verified end to end:

- `git show 30ac6962e -- Chart.yaml` → no version change
- prod pins `0.9.5`
- a live prod pod today: `terminationGracePeriodSeconds: 30`

So prod sandbox pods run a **30s** grace period while the shutdown git push alone self-limits at **30s** (`DEFAULT_GIT_TIMEOUT_MS`). That is precisely the "too tight, drops unsynced work" condition #4561 was written to fix, and it has been live the whole time. On spot reclaim / rollout / idle eviction, the push can be SIGKILLed mid-flight and the pod's `/app` emptyDir is the only other copy.

I'm not touching `deco-apps-cd` here — flagging it because it's user-data-loss shaped and it needs a decision on how far to jump.

**Relevance to #5339:** the flush there is dispatched before the git sync and awaited after, bounded at 3s, so it only consumes budget on a path that already finished the push. It doesn't make this worse. But it was sized against a grace period I'd assumed was 90.

pedrofrxncx:
Retargeting: the metrics this chart knob enables moved to a log line in #5346, which needs no endpoint and no egress decision. This PR now only matters if we want OTLP push from sandbox pods as a platform capability — see the discussion on #5339, which is in draft pending that. Holding here rather than merging a knob nothing sets.

pedrofrxncx:

[studio-bot:stale-close]

PR #5400, opened 2026-07-29
feat(sandbox): auto-save sandbox work to the remote

pedrofrxncx:
## Summary

The daemon only synced the sandbox working tree in `shutdown()`. Any ungraceful exit — SIGKILL on pod eviction, OOM, node loss, grace period elapsing before the push finishes — took the agent's work with it.

This adds `AutoCommitter` (`daemon/git/auto-commit.ts`): every 30s, if the tree is dirty or the branch has unpushed commits, commit and push. **On by default**, opt out with `git.autoCommit: false` in the daemon config (validated, works via `PUT /_sandbox/config` or a tenant-committed `.decocms/daemon.json`).

### Reuse over reinvention

`publish()` is split into its local half and its network half so the loop shares the rules instead of doing a second blind `add`/`commit`/`push`:

- `stageAndCommit()` — staging + commit, incl. the protected-branch refusal, the invalid-decofile-block net (`onInvalidBlock: "skip"`, same as the shutdown sync) and the org-fs `.git/info/exclude` filtering.
- `ensureOriginPushable()` — credentialed origin + the tokenless-GitHub guard.
- `publish()` is now those two plus `pushBranch()`; behavior unchanged.
- `pushBranchAsync()` — new: the loop's push goes through `gitAsync` so it never blocks the daemon's single event loop long enough to miss a health probe (Studio tears a sandbox down on one missed probe). Local git calls stay sync, matching `BranchStatusMonitor`'s existing 3s poll.

### Behavior notes

- Never pushes to a protected branch (checked before the commit, so no stray commit either) and never force-pushes — an autosave must not clobber a diverged `origin/<branch>`, same reasoning as the shutdown sync. A rejected push is left for the interactive publish to reconcile.
- Failures are logged and retried on the next tick; the shutdown publish remains the backstop.
- The trigger reads `BranchStatusMonitor`'s already-computed meta, so a clean tree costs zero git calls, and boot-dirty dev-server noise doesn't *trigger* a tick. The commit itself still stages everything dirty (identical to the shutdown sync) — carrying a little noise beats dropping the user's work.
- `unpushed > 0` is a trigger too, so commits the agent made itself via `bash` also get pushed.
- Overlapping ticks are latched, and `shutdown()` stops the loop before its own publish so the two can't race for `index.lock`.

## Testing

- New `daemon/git/auto-commit.test.ts`: `shouldAutoCommit` decision table, plus two real-git tests against a bare `file://` origin — a dirty tree ends up on the remote branch, and a disabled committer touches nothing.
- `bun test packages/sandbox/` → 878 pass / 0 fail, including the daemon e2e suite booting the rebuilt binary with the loop wired.
- `bun run check`, `bun run lint` (0 errors), `bun run fmt`, `knip` clean.

## Skipped

No Studio-side UI/org-flag to toggle it — the daemon config field is the switch; add a surface when someone asks for one.

---
## Summary by cubic
Auto-save sandbox work to the remote a few seconds after repo writes settle, with a 30s floor, to prevent data loss on ungraceful exits. Enabled by default, never pushes protected branches, and can be disabled with `git.autoCommit: false`.

- **New Features**
  - Added `AutoCommitter` that saves on change (5s trailing debounce) and on a 30s interval; commits and pushes when the tree is dirty or has unpushed commits, using async push to keep the daemon responsive.
  - Config toggle `git.autoCommit` (validated) via `PUT /_sandbox/config` or a tenant `.decocms/daemon.json`.
  - Skips protected branches and never force-pushes; logs failures and retries; shutdown publish remains the backstop.

- **Refactors**
  - Split `publish()` into `stageAndCommit` and `ensureOriginPushable` to reuse staging rules; added `pushBranchAsync` for non-blocking pushes.
  - Integrated start/stop into the daemon lifecycle, wired to the repo write funnel, and latched overlapping ticks; reuses `BranchStatusMonitor` metadata to avoid extra git calls.
  - Updated README and added tests for decision logic, debounce behavior, and real git flow.

<sup>Written for commit f474f81c4e28fe56f86e975c95541f97bf583ad7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5400?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/README.md
+++ b/packages/sandbox/README.md
@@ -31,6 +31,10 @@ never becomes another user's execution context.
 - Provision or attach to isolated execution environments.
 - Build and run the authenticated Bun daemon inside each environment.
 - Expose asynchronous filesystem, Git, task, terminal, and preview operations.
+- Auto-save work in progress to the remote every 30 s so an ungraceful exit
+  (SIGKILL on eviction / OOM / node loss) can't lose it. On by default; disable
+  per sandbox with `git.autoCommit: false` in the daemon config. Never pushes to
+  a protected branch.
 - Carry agent dispatch streams between Studio, the daemon, and a harness.
 - Proxy HTTP and WebSocket traffic to development servers inside a sandbox.
 - Support organization filesystem mounts and cleanup.
--- a/packages/sandbox/README.md
+++ b/packages/sandbox/README.md
@@ -31,10 +31,11 @@ never becomes another user's execution context.
 - Provision or attach to isolated execution environments.
 - Build and run the authenticated Bun daemon inside each environment.
 - Expose asynchronous filesystem, Git, task, terminal, and preview operations.
-- Auto-save work in progress to the remote every 30 s so an ungraceful exit
-  (SIGKILL on eviction / OOM / node loss) can't lose it. On by default; disable
-  per sandbox with `git.autoCommit: false` in the daemon config. Never pushes to
-  a protected branch.
+- Auto-save work in progress to the remote — a few seconds after repo writes
+  settle, and at worst every 30 s — so an ungraceful exit (SIGKILL on eviction /
+  OOM / node loss) can't lose it. On by default; disable per sandbox with
+  `git.autoCommit: false` in the daemon config. Never pushes to a protected
+  branch.
 - Carry agent dispatch streams between Studio, the daemon, and a harness.
 - Proxy HTTP and WebSocket traffic to development servers inside a sandbox.
 - Support organization filesystem mounts and cleanup.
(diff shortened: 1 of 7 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #5524, opened 2026-07-31
fix(web): don't strand a second account in the previous user's org

pedrofrxncx:
## Problem

Signing in with a different account on a browser that already used Studio lands on **No access** instead of that account's own org:

> `studio.decocms.com/ferragens-floresta` → "You don't have access to Ferragens Floresta."

`studio:last-org-slug` and `studio:last-location` are browser-global, but they describe **one principal's** history. A second account (or a session that expired and was replaced) inherits them, and cold entry (`/`) redirects straight into an org it can't access.

The gate's existing self-heal didn't catch it: it bounced back to `/` only when `lastOrgSlug` still matched the org, while the redirect that actually lands users there now comes from `lastLocation` (checked first in `homeRoute`) — and `orgLayout.beforeLoad` overwrites `lastLocation` with the current org on arrival, so it can't answer "did we send them here?" either.

## Changes

- **Principal-scope the restore state.** `studio:last-user-id` records who owns the two keys; the shell calls `claimRestoreStateFor(userId)` as soon as the session resolves and wipes them if it's somebody else. Same pattern as the already user-scoped org cache (`readCachedOrg`). Sign-out clears them too (`clearPersistedQueryCache`, whose contract is already "next user starts clean").
- **Fix the recovery.** `homeRoute` marks its restore-driven redirects (per-tab `sessionStorage`); `OrgAccessGate` reads the marker. A restored org the user can't access → back to `/`, which now resolves their default org. A deliberate visit to a foreign org (shared link) still shows the screen with a working "Go to home".
- **Clear both keys**, not just the slug, when landing on an archived org or deleting the org you're in — `lastLocation` pointing at either bounced the user right back.

## Testing

- `apps/web/src/lib/last-location.test.ts` — principal scoping, marker matching, and that a re-render gets the same marker answer (so StrictMode can't turn a bounce into the dead-end screen).
- `bun test apps/web/src/lib/` green (195 pass); full `bun test apps/web` has fewer failures than `main` in this environment (61 vs 68 — the pre-existing ones are unrelated).
- `bun run fmt`, `bun run lint` (0 errors), `bun run check` pass.

The `query-persist` test's window stub moved to `Object.defineProperty`: happy-dom makes `window.localStorage` a readonly accessor, so the plain assignment threw whenever another test file registered happy-dom first.

## Manual check

Sign in as A, visit an org only A can access, sign out, sign in as B → B lands on B's own org.

---
## Summary by cubic
Fixes “No access” loops when signing in with a different account by scoping restore state to the current user and tracking restore-driven redirects. Also clears stale state on sign-out, org deletion, and archived orgs to avoid bouncing back.

- **Bug Fixes**
  - Principal-scoped restore state with `studio:last-user-id`; `claimRestoreStateFor(userId)` clears stale `lastLocation`/`lastOrgSlug` as soon as the session resolves. Sign-out also clears it via `clearPersistedQueryCache`.
  - `homeRoute` marks restore redirects per-tab (`markRestoreRedirect`), and `OrgAccessGate` consumes them (`consumeRestoreRedirect`) to send users back to `/` only for restore-driven arrivals without access; direct shared links still show the gate.
  - Replaced ad-hoc slug clearing with `clearRestoreState()` when deleting an org or landing on an archived org.
  - Added tests for principal scoping and the per-tab marker (stable across re-reads); fixed `window.localStorage` stubbing using `Object.defineProperty`.

<sup>Written for commit 82eac493d90fe253f9213fd509bccb01050d8602. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/web/src/components/org-access-gate.tsx
+++ b/apps/web/src/components/org-access-gate.tsx
@@ -4,7 +4,11 @@ import { PendingInviteScreen } from "@/components/pending-invite-screen";
 import { RequestPendingScreen } from "@/components/request-pending-screen";
 import { RequestToJoinScreen } from "@/components/request-to-join-screen";
 import { useOrgAccessStatus } from "@/hooks/use-org-access-status";
-import { clearLastLocation, readLastLocation } from "@/lib/last-location";
+import {
+  clearLastLocation,
+  consumeRestoreRedirect,
+  readLastLocation,
+} from "@/lib/last-location";
 import { LOCALSTORAGE_KEYS } from "@/lib/localstorage-keys";
 
 /**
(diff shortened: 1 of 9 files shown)


viktormarinho:
curious, instead of doing this:

```
 lastOrgSlug: () => `studio:last-org-slug`,
 lastLocation: () => `studio:last-location`,
 /** Principal that owns the two keys above (see claimRestoreStateFor). */
 lastUserId: () => `studio:last-user-id`,
```
  
Couldn't we just remove that "lastUserId" and key by user id the others making me keep caches for all accounts i have?

pedrofrxncx:

[studio-bot:stale-close]

PR #5589, opened 2026-08-03
feat(thread): let any org member answer a thread awaiting input

guitavano:
## Summary

When a **QA Agent** or **Code Reviewer** thread pauses on a `user_ask` (status `requires_action` → "Requer entrada"), only the person who started the task could respond. Reviewer threads are spawned owned by the human who created/assigned the task (`created_by = task.assignedBy ?? task.createdBy`), so every other org member hit a read-only composer and the run was stuck on one person.

This adds a **single, scoped exception**: while a thread is in `requires_action`, **any authenticated org member may answer**. Outside that state, non-owners stay read-only exactly as before.

## Changes

- **`packages/shared/src/thread/access.ts`** — new pure helper `canRespondToThread({ createdBy, userId, status })`, the single source of truth for the rule, used by both gates. Unit-tested (`access.test.ts`).
- **`apps/api/.../dispatch-run.ts`** — the authoritative write gate now defers to the helper. When a non-owner answers a `requires_action` thread, `created_by` (ownership) is **unchanged**; the responder is recorded in `updated_by` (best-effort, never blocks the run) so the trail isn't lost.
- **`apps/web/.../chat/input.tsx`** — the read-only composer gate defers to the same helper, so a teammate viewing a thread awaiting input gets a working composer.

## Why it's safe

Org membership is already enforced upstream (org-scoped routes + org-scoped thread fetch), and thread **reads** are already open to any org member (`validateThreadAccess` — *"Does NOT enforce ownership"*). This only aligns the *write* path with existing read semantics, and only in the awaiting-input state. As soon as a teammate answers, the thread flips to `in_progress` and the composer returns to read-only for non-owners.

## Testing

- `bun test packages/shared/src/thread/access.test.ts` — ✅ 4 pass (owner always allowed; non-owner allowed only on `requires_action`; read-only for every other status; permissive fallback on missing ids).
- `bun run check` (shared / api / web) — ✅ type-checks clean.
- `bun run lint` — ✅ 0 errors on changed files.
- Not covered by e2e: reaching `requires_action` legitimately needs a real agent run with a `user_ask`, which requires a model provider not present in the test org (per TESTING.md, e2e must not depend on an absent provider tier) — so the decision logic is covered at the unit tier, its correct home.

---
## Summary by cubic
Allow any org member to answer a thread paused for input (`requires_action`) so QA Agent/Code Reviewer runs aren’t blocked. Enforcement now happens at API ingress to read the correct status before dispatch.

- **New Features**
  - Added `canRespondToThread` in `@decocms/shared/thread/access` as the single rule source, with unit tests.
  - Web composer (`apps/web/src/components/chat/input.tsx`) uses the helper to enable input during `requires_action`.
  - Ownership stays with the creator; non-owner responders are recorded in `updated_by`. Access remains org-scoped.

- **Bug Fixes**
  - Moved write-authorization to `validate()` in `apps/api/src/api/routes/decopilot/routes.ts` (pre-dispatch) to avoid reading a post-fence `in_progress` status and wrongly rejecting teammates.
  - Removed the post-fence throw; `apps/api/src/api/routes/decopilot/dispatch-run.ts` now only stamps `updated_by` best-effort.
  - Added e2e coverage: `packages/e2e/tests/thread-requires-action-access.spec.ts` validates deny/allow behavior around `requires_action`.

<sup>Written for commit 369fcccdec065089a106bafae6609b24bea07f86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


guitavano:
## Self-review (multi-perspective) — found & fixed a blocker

Ran a parallel review (correctness/concurrency, security, testing, architecture). Summary and the fix (2nd commit):

**Blocker (fixed):** The gate was in `prepareRun` (dispatch-run.ts), but the dispatch gate mints a run fence *before* `prepareRun`, and `setRunFence` flips the thread to `in_progress` (`threads.ts:928`, called at `thread-gate-workflow.ts:212`). So the check always read `in_progress`, never `requires_action` → **every teammate response was rejected**, and the thread was left stuck `in_progress`. The pure unit test passed but proved nothing about the wired path.

**Fix:** Moved the authorization to request **ingress** in `validate()` (POST /messages), where the pre-dispatch `threads.status` and the authenticated caller are both available — before the message is persisted and before the fence is minted. This also fixes a pre-existing issue where a rejected non-owner write had already persisted its user message. POST /messages is the only caller-identity dispatch path (background tools / task-board recovery all run as the owner), so removing the now-redundant post-fence check widens nothing.

**Added the wiring test that would have caught it:** `packages/e2e/tests/thread-requires-action-access.spec.ts` — owner creates a thread, a teammate joins the org, status is seeded via DB; asserts the teammate is **denied** on a settled thread and **allowed** once it awaits input (same member, same thread, only status differs). Also added empty-string-id cases to the unit test.

**Verified clean (verdicts otherwise):** org-scoping ✓, model-permission still fails closed for the responder's role ✓, `updated_by` not clobbered ✓, control actions (cancel/stop) stay owner-only ✓. `updated_by` wording softened to "audit" (it's not surfaced in any UI yet).

**Not run locally:** the new e2e needs Postgres; this environment has none (the API dev server's `migrate` step can't reach 5432), so it type-checks/lints clean but was not executed here — CI will run it. Unit tests + `bun run check` (shared/api/web/e2e) + `bun run lint` all pass.

PR #5601, opened 2026-08-03
feat(reports): mount report-pages under /api/report (og.png without touching nginx)

viktormarinho:
## Context

The /report/<site> SEO/OG merged in #5588 doesn't show in production. It's not cache: the nginx front door only proxies `^/(api|mcp|oauth-proxy|\.well-known|org|health|metrics)` to the Bun API — `/report` falls through to the static SPA fallback, so `createReportPagesRoutes` never runs. Alternative to changing nginx (PR #5600, closed).

## What changes

- `createReportPagesRoutes` is mounted **twice**: `/report` (as before) and `/api/report` — the latter goes through the front door today, no nginx image rebuild. `report` is already a reserved organization slug, so it can't collide with the `/api/:org` catch-all.
- `og:image` in the head now points at `${origin}/api/report/:domain/og.png` (the `/report/...` path would never reach this server).

Verification: `tsc` and `oxlint` clean; scoped `bun test` (app + routes) has an **identical** failure set to main (local-infra tests), empty diff.

## What's STILL missing for full SEO

The per-report head (title with score, description, og:image) is only emitted when the `/report/:domain` **HTML** reaches this server — and nginx keeps serving the static SPA on that path. Since the front door doesn't change here, close the loop with a **Cloudflare Transform Rule** (studio.decocms.com zone): path rewrite `^/report/(.+)` → `/api/report/$1`. It's origin-side — the browser URL stays `/report/<site>` — and the handler already emits `canonical`/`og:url` pointing at `/report/<site>`. Without it, unfurlers keep seeing the SPA's default meta.

### DIFF
--- a/apps/api/src/api/app.ts
+++ b/apps/api/src/api/app.ts
@@ -1295,7 +1295,13 @@ export async function createApp(options: CreateAppOptions = {}) {
 
   // Auth-gated report page + domain-derived metadata. API-only/test apps safely
   // return 404 for the HTML shell when no built client directory is supplied.
-  app.route("/report", createReportPagesRoutes(options.clientDir));
+  // Mounted twice: the nginx front door only proxies /api-prefixed paths to
+  // this server (everything else falls through to the static SPA), so the
+  // share-card image URL (og:image → /api/report/:domain/og.png) rides the
+  // /api mount to stay reachable in production without touching nginx.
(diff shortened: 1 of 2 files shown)


viktormarinho:
Closing — going with the nginx front-door fix instead (route /report to the Bun API), which serves the full per-report head, not just the image.

PR #5608, opened 2026-08-03
fix(sandbox): cap the Go daemon's plain-text read path at 10MB

pedrofrxncx:
**Source**: bug found while auditing `packages/sandbox/daemon-go/internal/routes/fs.go` for the sandbox daemon focus area. A prior attempt at this same fix (on the now-deleted TS daemon) was blocked pre-open for an unrelated oversized diff — this is a small, isolated fix on the Go daemon, which is now the only daemon.

**Why a maintainer wants this**: the `/read` route caps images at 5MB (`maxImageBytes`) and file transfers at 500MB (`maxTransferBytes`), but the plain-text branch has no cap at all — `os.ReadFile` loads the whole file, then `strings.Split` and the response `strings.Builder` each hold another full copy. A large text file anywhere in the repo (a generated log, a committed fixture, a build artifact) turns one `/read` tool call into a multi-hundred-MB memory spike. Per the repo's own docs, Studio marks a sandbox **dead on a single missed health probe** and tears the pod down — so an OOM here doesn't just fail one call, it kills the sandbox.

**Failure scenario**: any tool/agent calls `/read` on a text file larger than a few hundred MB sitting in the sandbox's working tree → the daemon reads it into memory 2-3x over → OOM → health probe miss → Studio tears the sandbox down mid-session.

**Fix**: added a `maxTextBytes = 10 * 1024 * 1024` cap (matching the daemon's existing `installLogMaxBytes` convention) checked via `stat.Size()` before the `os.ReadFile` call, returning the same shaped 400 error the image-size cap already uses. Regression test (`fs_read_test.go`) exercises both the oversized-file rejection and that an under-cap file still reads fine.

**Reviewer command**: `cd packages/sandbox/daemon-go && go test ./internal/routes/... -run TestRead -v`

**Locally verified**: `go build ./...`, `go test ./...` (full package, all green), `go vet ./internal/routes/...`, `gofmt -l` (clean) on both changed files. Full CI validates the rest.

---
## Summary by cubic
Cap the sandbox daemon’s `/read` plain-text path at 10MB to prevent large-file OOMs that can kill the sandbox. Adds a size check before reading and returns a 400 error for oversized files.

- **Bug Fixes**
  - Add `maxTextBytes = 10 * 1024 * 1024` in `packages/sandbox/daemon-go/internal/routes/fs.go`; reject files over cap with 400.
  - Check `stat.Size()` before `os.ReadFile` to avoid large allocations.
  - Add `fs_read_test.go` tests for oversized rejection and an allowed under-cap read.

<sup>Written for commit bb157ba2514d6134b0b7c1a1854da1565ed27072. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/daemon-go/internal/routes/fs.go
+++ b/packages/sandbox/daemon-go/internal/routes/fs.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	maxImageBytes    = 5 * 1024 * 1024
+	maxTextBytes     = 10 * 1024 * 1024
 	maxTransferBytes = 500 * 1024 * 1024
 	transferDeadline = 5 * time.Minute
 )
@@ -131,6 +132,11 @@ func Read(deps FsDeps) http.HandlerFunc {
 
 		if bytes.IndexByte(probe, 0) >= 0 {
 			httpx.Error(w, 400, "File appears to be binary and is not a supported image format (jpeg/png/gif/webp).")
+			return
+		}
+
+		if stat.Size() > maxTextBytes {
+			httpx.Error(w, 400, fmt.Sprintf("File too large (%d bytes; cap is %d)", stat.Size(), maxTextBytes))
 			return
 		}
 
--- a/packages/sandbox/daemon-go/internal/routes/fs_read_test.go
+++ b/packages/sandbox/daemon-go/internal/routes/fs_read_test.go
@@ -1,0 +1,62 @@
+package routes
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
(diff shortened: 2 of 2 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #5733, opened 2026-08-05
fix(local-mode): persist auth secret so bunx decocms survives restarts

guitavano:
## Summary

A brand-new user who just runs `bunx decocms` gets trapped in an **unbreakable login loop** on the second launch. This makes zero-config local mode "just work" across restarts.

### Root cause

Local mode ships without `BETTER_AUTH_SECRET`, so Better Auth mints a **fresh random secret every process start** (`apps/api/src/auth/index.ts:458`). Each restart:
- invalidates existing session cookies, and — worse —
- fails to decrypt the JWKS private key persisted on the previous boot:

```
BetterAuthError: Failed to decrypt private key. Make sure the secret currently
in use is the same as the one used to encrypt the private key.
[5xx Response] GET /api/auth/get-session - 500
```

`get-session` 500s → the UI bounces back to login → loop, with no way out short of wiping `~/deco` or manually pinning an env var.

### Fix

1. **`settings/local-secret.ts` (new)** — in local mode, generate a random secret once, persist it to `~/deco/local-auth-secret` (mode `0600`), and reuse it across restarts.
2. **`settings/pipeline.ts`** — when `localMode && !BETTER_AUTH_SECRET`, resolve the persisted secret. Server deployments are untouched (they set the env explicitly; the empty value falls through to the existing random-per-process behavior).
3. **`auth/local-mode.ts`** — decouple the local admin password from `betterAuthSecret` (now a fixed constant). Without this, the newly-persisted random secret would no longer match the password hash of an **existing** local DB seeded under the old fallback, breaking auto-login on upgrade. Local mode is loopback-only auto-login, so a constant is safe; seed and sign-in read the same value.

## Testing

- New unit test `settings/local-secret.test.ts` (generates, is stable across calls, creates the data dir, writes `0600`) — passes
- `bun run --cwd apps/api check` (typecheck) — passes
- `bun test apps/api/src/auth/` (77 tests) — passes
- `bun run fmt` / `bun run lint` — 0 errors

## Affected areas

`bunx decocms` / CLI local mode (`DECOCMS_LOCAL_MODE=true`) only. The root `bun run dev` workflow (loads `.env`) and server deployments are unchanged.

---
## Summary by cubic
Persist and propagate a stable auth secret in local mode to stop the login loop across restarts and dev child processes. This lets `bunx decocms` work across restarts with no env setup; server deployments are unchanged.

- **Bug Fixes**
  - Persist a random Better Auth secret under the data dir (0600) and reuse it across restarts to keep cookies valid and decrypt the JWKS key.
  - In `settings/pipeline.ts`, when `localMode` and no `BETTER_AUTH_SECRET`, load the persisted secret; in `cli/commands/dev.ts` forward it to the spawned dev child as `BETTER_AUTH_SECRET`. Production still uses explicit env.
  - Use a fixed local admin password (no longer tied to the secret) to keep auto-login on existing local DBs; also remove unused docs components (`Step`, `StepGroup`) to keep lint green.

<sup>Written for commit f90420b85d08b1379c12b23e07aa05b31f79d0bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5733?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tlgimenes (approved):

guitavano:
## Self-review (parallel multi-perspective critique) — 1 gap found & fixed

Ran a parallel review (correctness/architecture + security/testing critics). Both **independently** flagged the same real gap, now fixed in `aeead7db9`:

### Fixed — `bunx decocms dev` / `bun run dev` still looped
The first commit fixed only the **serve** path (in-process). The **dev** path spawns a `dev:servers` child that re-derives Settings via `initSettingsFromEnv()` — which does *not* run local-secret persistence — and the child env block never forwarded `BETTER_AUTH_SECRET`. So the child minted a fresh random secret every boot, leaving the loop intact on the dev path. Fixed by mirroring the parent's resolved secret into the child env (same convention as `DATABASE_URL`/NATS/S3).

### Verified safe (checked, no change needed)
- **Serve path**: `buildSettings()` → `setGlobalSettings()` runs before `import("../../index")`, so `auth/index.ts`'s module-level `authSecret` reads the persisted value. No reader sees the pre-override empty value.
- **dataDir**: cwd-independent, stable across restarts; `mkdirSync(recursive)` before write is sufficient.
- **`getLocalAdminPassword` constant**: both callers (seed + loopback-guarded sign-in) agree; the zero-config cohort is unaffected (old empty secret already yielded `"local-mode-default"`).
- **Plaintext `0600` secret file**: consistent with existing repo precedent (NATS `keys.json`/`cluster.creds` persist the same way under dataDir).

### Known minor limitations (deferred, documented)
- **Concurrent first-run race**: two simultaneous fresh-dataDir starts could generate divergent secrets. Low likelihood (they contend on embedded Postgres too) and the dev parent writes the file *before* spawning the child (sequential), so not a real path. Not worth a lock.
- **Existing DB seeded under a custom `BETTER_AUTH_SECRET`**: auto-login breaks for that narrow cohort (uncommon; pure `bunx decocms` had an empty secret, so unaffected).
- **Testing**: e2e "restart preserves local session" coverage doesn't exist yet and is a heavier new harness — recommend as a follow-up rather than blocking this focused fix.

PR #5824, opened 2026-08-06
feat(sandbox): reimplement the L2 golden tier in the Go daemon

nicacioliveira:
The cross-node golden tier has no implementation. This restores one.

## What happened

The tier was written in TypeScript (#5351, #5399) and **deleted with that daemon** in #5575, before it ever ran in a deployed config. Nothing replaced it — the Go daemon logged

```
GOLDEN_CACHE_REMOTE is set but this daemon implements the node-local golden tier only
```

and installed from scratch. Meanwhile the infrastructure kept being built around it: the bucket (decocms/terraform-eks-cluster#88), the PV/PVC ownership refactor (#92, decocms/studio#5806) and the rollout (decocms/deco-apps-cd#203). So prod today has the volume mounted at `/golden-cache` with nothing on either end of it. The `RestoreL2` metric constant in `depmetrics.go` survived too — only the code was gone.

## What this does

Ports restore, publish and prune to Go, keyed **identically** to L1 (same credential-stripped repo hash, same lockfile hash) so the two tiers can never disagree about what a key means. A test asserts credentials do not change the key and that two repos never collide — bun does not re-verify cache content, so per-repo isolation is the boundary both tiers rely on.

**Orchestrator wiring:** L1 reflink → L2 extract → install. An L2 hit leaves `pendingGolden` set, so the healthy-boot transition seeds *this* node's L1 from the extracted tree and the next pod here gets the reflink instead of another extract. That same transition runs the L2 publish, which no-ops on an existing key, so the two are safe together.

## Two deliberate departures from the original

**The pipe is `os.Pipe`, not `bash -c` with `pipefail`.** Both children get one real kernel fd pair, so a multi-GB tree never traverses this process. That matters twice: the health probe must keep answering (Studio tears the pod down on a single miss), and a *relayed* copy is exactly what truncated this transfer in the TS version — using `os/exec`'s `StdoutPipe` here would reintroduce that relay. It also drops the bash dependency and the shell quoting it required.

**No shell means no `pipefail`,** so both exit codes are checked explicitly — consumer first, because its stderr is the line that explains a corrupt archive. `tar exit 2` alone is unactionable; it took a hand-run of the pipe to learn it meant "Unexpected EOF in archive".

## Kept, because they are load-bearing

- **An archive, never the tree.** Per-operation metadata latency is fatal across ~100k files, a non-issue for one sequential blob. The per-file cost is paid by the local extract.
- **Extract into staging, then rename.** An interrupted tar would otherwise leave a partial `node_modules` that later code reads as complete — the boot skips install and fails somewhere unrelated.
- **Write straight to the final key, not temp+rename.** A blob store has no rename and does not need one: an object becomes visible only once its upload completes, which is the same guarantee. A POSIX store would need it back, and the comment says so.
- **Read the archive back before reporting success.** Publish no-ops on an existing key, so a corrupt object would make every node in the fleet miss forever with nothing to repair it.

`zstd` is already in the sandbox image (`packages/sandbox/image/Dockerfile:42`) — it survived the TS deletion. `main.go` now warns when it is *missing* instead of warning that the tier is unimplemented.

## Testing

Real files, real `tar`/`zstd`, no mocks:

| test | why it exists |
|---|---|
| publish on one root, restore on another | the point of the tier: a cold root with no `node_modules` gets a complete tree, contents verified |
| ~8 MB payload | past any pipe buffer — the case that caught a truncating implementation in TS, where a 1 KB fixture passed |
| dormant when unconfigured | discriminating counterpart; without it the two above would pass even if the code ignored its configuration |
| truncated archive | fails closed: no partial tree, no staging left behind |
| corrupt (never-valid) archive | same, for bytes that were never an archive |
| idempotent publish | a second publish does not rewrite a live key — which matters because there is no rename to hide a partial object behind |
| prune | TTL, per-repo cap, non-archive entries left alone, repos independent |
| `.vite` excluded | pod-local churn must not travel in a shared archive |

`go build ./...`, `go vet`, `gofmt` clean; the full `internal/setup` suite passes. Tests skip rather than fail when `zstd` is absent, so a contributor without it does not get a red suite that trains them to ignore it.

**Not covered:** publish's read-back guard. Publish no-ops on an existing key so a corrupt object cannot be planted ahead of it, and making `zstd` emit a bad archive on demand would mean faking the pipe — which defeats testing against real tools. Noted in the test file rather than papered over.

## Not validated on a real pod, and it cannot be yet

The tenant mount is **read-only** by design, so publish always fails there and the store stays empty. Restore needs something to have published first.

So this PR makes the tier *possible*, not *live*: with a seeded store, sandboxes will restore from it and `sandbox.deps.restore{source="l2"}` starts reporting. The remaining piece is a trusted seeder holding the writable mount — deliberately not a sidecar in the sandbox pod, since a package manager installs cached content as-is and write access to a shared store would turn a node-local RCE surface into a fleet-wide one.

Per the repo norm, "done" means a tailed pod showing an L2 hit. This is not that yet.

---
## Summary by cubic
Reimplements the cross-node L2 golden cache in the Go daemon to restore, publish, and prune shared `tar`+`zstd` archives, cutting cold-node installs and wiring L2 before install in the orchestrator.

- **New Features**
  - L2 restore/publish/prune keyed exactly like L1 (credential-stripped repo hash + lockfile hash).
  - Orchestrator tries L2 before install; on hit, seeds this node’s L1, emits `sandbox.deps.restore{source="l2"}`, marks install success; after healthy boot it runs L2 publish and prune.
  - Restore extracts to staging then renames; no partial trees on failure.
  - Publish is idempotent, writes directly to the final key, validates by reading back the archive; excludes pod-local caches like `.vite`.
  - Prune by TTL and per-repo cap; safe post-publish; no-ops on read-only mounts.
  - Robust streaming via `os.Pipe` with explicit exit checks; large streams never traverse the daemon process.
  - Dormant unless `GOLDEN_CACHE_REMOTE` is set; `main.go` warns when `zstd` is missing.

- **Migration**
  - Seed the shared store from a trusted publisher with a writable mount; tenant pods stay read-only. Once seeded, sandboxes restore from L2 automatically or fall back to install.

<sup>Written for commit 2f539d6edc011f19d28f6b5fb3ebcf614a9fa85a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5824?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/daemon-go/internal/setup/orchestrator.go
+++ b/packages/sandbox/daemon-go/internal/setup/orchestrator.go
@@ -146,6 +146,12 @@ func (o *Orchestrator) PublishPendingGolden() {
 	pending.Log = func(m string) { o.rawChunk(m + "\r\n") }
 	PublishGolden(*pending)
 	PruneGoldens("")
+	// L2 last: it is the slow one (compress + read-back over the network), and a
+	// no-op both when the key already exists — the common case after an L2
+	// restore — and when GOLDEN_CACHE_REMOTE is unset. The store is read-only in
+	// a tenant pod, so today this only ever logs a failure to write; that changes
+	// when a trusted publisher gets the writable mount.
+	PublishRemoteGolden(RemoteGoldenFrom(*pending))
 }
 
 func (o *Orchestrator) enqueue(step Step) {
@@ -420,6 +426,23 @@ func (o *Orchestrator) stepInstallInner() bool {
 		o.pendingGolden = nil // restored an existing golden — nothing to publish
 		o.mu.Unlock()
 		EmitDepsRestore(RestoreL1, cloneUrl, elapsedMs(), bootId)
+		o.markInstallSucceeded(cfg)
+		return true
+	}
+
+	// L2: the same lockfile may already be archived on the shared store even
+	// though this node has never built it. Extracting it beats installing, and it
+	// is the only tier that helps a pod that landed in a cold zone.
+	//
+	// A hit leaves pendingGolden set, so the healthy-boot transition seeds THIS
+	// node's L1 from the extracted tree — the next pod here gets the reflink
+	// instead of another extract. PublishRemoteGolden no-ops on an existing key,
+	// so it is safe that the same transition also runs the L2 publish.
+	if TryRestoreRemoteGolden(RemoteGoldenFrom(golden)) {
+		o.mu.Lock()
+		o.pendingGolden = &golden
+		o.mu.Unlock()
+		EmitDepsRestore(RestoreL2, cloneUrl, elapsedMs(), bootId)
 		o.markInstallSucceeded(cfg)
 		return true
 	}
--- a/packages/sandbox/daemon-go/internal/setup/remotegolden.go
+++ b/packages/sandbox/daemon-go/internal/setup/remotegolden.go
@@ -1,0 +1,394 @@
+package setup
+
+// L2: the cross-node golden tier.
+//
+// The node-local golden (golden.go) only hits when the pod lands on a node
+// already warm for its repo. The sandbox pool churns ~170 nodes a day across
+// three AZs with ~80% fresh pods, so a large share of boots land somewhere cold
+// and pay a full install, and cross-zone a node-local cache cannot help at all.
+// This tier removes the same-node dependency: one archive per
+// (repo, pm, lockfile) on a shared store, restorable on ANY node in ANY zone.
+//
+// ONE HARD RULE: a compressed ARCHIVE on the shared store, never the
+// node_modules TREE. A shared store charges per-operation metadata latency —
+// fatal across ~100k small files, a non-issue for a single large sequential
+// blob. The per-file cost is paid by the local extract, on local disk. Mounting
+// a tree here would be slower than no cache at all.
+//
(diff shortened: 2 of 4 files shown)


nicacioliveira:
**Merge dependency:** #5843 is stacked on this branch and adds the *publish* half — a node-level DaemonSet that fills the shared store.

Neither PR is useful alone. This one restores from a store nothing fills (the tenant mount is read-only by design, so no sandbox boot can publish); #5843 fills a store nothing reads. **Merge both or neither** — merging only this one leaves prod with the volume mounted, the env var set, and a store that stays permanently empty, which is the exact state that let the tier sit broken and unnoticed for three months after #5575.

#5843 also adds the org to the archive key, which changes the key shape. Worth landing together rather than reshaping keys after anything is published.

pedrofrxncx (approved):

PR #5836, opened 2026-08-07
feat(sandbox-env): per-tenant pod resources for tenant warm pools

pedrofrxncx:
Follow-up to #5830. A tenant whose repo needs more memory than the fleet default
had no way to ask for it: `resources` is a pod-spec field, so it can be neither
per-claim nor per-pool — it needs its own `SandboxTemplate`.

## Shape

- The sandbox pod spec moves into a shared partial
  (`sandbox-env.sandboxTemplateSpec`). Setting **`tenantPools[].resources`**
  renders a second `SandboxTemplate`, named after the pool, from that **same**
  partial — so a hand-copied 400-line duplicate can't drift the first time
  someone edits one of them — and points the pool at it.
- Pools without `resources` keep sharing the chart-wide template. Their render
  is unchanged.
- The claim now names **the template the pool's pods were actually built from**,
  which Studio reads back off the operator's own `SandboxWarmPool` on each
  reconcile tick. Deliberately not a second string in
  `STUDIO_SANDBOX_TENANT_POOLS`: a claim naming a different template than the
  pod never binds, and we've already been bitten twice this week by two sides
  having to agree on one literal.

```yaml
tenantPools:
  - name: tenant-acme-site-prod
    size: 2
    resources:
      requests: { cpu: "2", memory: "6Gi" }
      limits: { memory: "12Gi" }
```

## Verification

The refactor is the risky part, so it was gated on a render diff. Rendering the
chart with prod-shaped values before and after, the **only** difference is the
`helm.sh/chart` label from the version bump:

```
34c34
<     helm.sh/chart: sandbox-env-0.12.1
---
>     helm.sh/chart: sandbox-env-0.13.0
```

Nothing else moves — the shared SandboxTemplate is byte-identical, so existing
installs that set no `resources` see no spec change and their pools don't roll.

Also verified: with two pools, one overriding resources and one not, the
overriding pool renders its own template with the overridden limits and points
at it, while the plain pool still points at `studio-sandbox-<env>`.

## Not in this PR

- Per-tenant **node placement** (on-demand / `do-not-disrupt`). Same lever — it
  is now one `$pool.` reference away — but it wants its own values shape and a
  cost conversation.
- The `updateStrategy: OnReplenish` pin, which is #5836's branch; this PR stays
  off those lines to avoid conflicting with it.

---
## Summary by cubic
Adds per-tenant pod resources for tenant warm pools by rendering a pool-specific SandboxTemplate when `tenantPools[].resources` is set; pools without overrides keep the shared template. Claims now use the template the pool used (fallback to the shared template before first reconcile) to avoid bind failures.

- Helm: moves the sandbox pod spec into a shared partial (`sandbox-env.sandboxTemplateSpec`). Setting `tenantPools[].resources` renders a per-pool `SandboxTemplate` and points the pool’s `sandboxTemplateRef` at it; other pools keep the shared template (byte-identical render). Chart version bumped to `0.16.0`. `values.yaml` documents `tenantPools[].resources`.
- Studio/Operator: `listWarmPoolPods()` returns `{ templateRef, pods }`. The runner caches per-pool template refs and uses them in claims; it falls back to the shared template until the first reconcile.
- CI: updates `helm-test.yml` to render tenant pools with and without per-pool resources and assert two `SandboxTemplate`s (shared + overriding pool).

<sup>Written for commit c1de05991a29c74cc59c098b6f50a6d5b0a6036e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:

[studio-bot:stale-close]

PR #5843, opened 2026-08-07
feat(sandbox): node-level uploader that fills the shared golden store

nicacioliveira:
**Stacked on #5824** — that PR adds the L2 *restore*; this one adds the *publish*. Base is its branch, so the diff here is only this change and GitHub will retarget to `main` when #5824 merges.

Neither is useful alone: #5824 restores from a store nothing fills, and this fills a store nothing reads. **Merge both or neither.**

## The gap this closes

A sandbox pod's L2 mount is read-only by design, so nothing a tenant boots can publish. The store is therefore permanently empty, and #5824's restore always misses. This is the missing half.

## Why a DaemonSet and not a sidecar

Writing a store shared across nodes needs write access, and a sandbox runs untrusted code — a package manager installs cached content as-is, so a tenant able to write it could poison another node's boot. The tenant's write path stops at its own node's hostPath, which is *already* the accepted trust domain there (`depsCache`'s own comment: *"Do NOT widen to a shared cache without making it read-only to tenant pods"*). This carries goldens the rest of the way.

| | sidecar | DaemonSet |
|---|---|---|
| sandbox PodSpec | +1 container in **every** pod | untouched |
| packing density (just tuned in terraform#91) | reduced | unaffected |
| can degrade the health probe | yes | no |
| writable mount lives | in the tenant's pod | once per node, in infrastructure |
| compression CPU | competes with the user's dev server | capped (50m/500m), nobody waits on it |

Asserted against the rendered output, not just intended: the SandboxTemplate never references the writable claim, and the DaemonSet mounts the hostPath `readOnly: true`.

## Keyed by org — this is what makes fleet-wide publishing acceptable

A repo hash is derived from the clone URL, so two orgs cloning the same public template land on **one key**. Node-locally that is a single trust domain; fleet-wide it would let one org's tree restore into another's sandbox.

The daemon is the only process holding both facts at once — the org from Studio's config push, and the golden it writes — so it records them together beside the tree. A golden **without** a readable org is skipped rather than uploaded under an unscoped key: an archive whose owner is unknown must not exist.

`orgId` already reached `buildConfigPayload` via `EnsureOptions.tenant`; it was simply never put on the wire. Getting the key right *now* matters because adding the org later would change every key and invalidate the whole store.

## What it uploads is not configurable, and cannot be

Sandboxes are multi-tenant and run third-party code, so the set of `(repo, lockfile)` pairs is not knowable ahead of time. There is no list to curate and no image to pre-bake — measured: 18 distinct repos in a 41-event sample over 30 days, long tail, no concentration to exploit. It forwards what a real boot produced and the daemon already judged healthy enough to publish node-locally, and nothing else.

## Cost

One PUT per `(org, repo, lockfile)` across the whole fleet — an existing key is a HEAD and a skip. At ~300 MB/archive and ~200 lockfiles that is single-digit dollars a month in S3, and $0 transfer via the VPC Gateway Endpoint. The real cost is compression CPU on the sandbox NodePool, which is why the limit is low and sweeps are sequential.

## Shape

Same binary as the daemon, via a `golden-uploader` subcommand — the archive format lives in `internal/setup`, which Go's `internal/` rule keeps unimportable from outside the module, and reusing the sandbox image means no second artifact to build or keep in lockstep. It is **not** the daemon: no HTTP server, no lifecycle, no tenant.

Chart `0.12.0`, `goldenUploader.enabled` default **off**.

## Testing

Real files, real `tar`/`zstd`, no mocks. 26 tests pass in `internal/setup`; `go build`, `go vet`, `gofmt` clean; `helm lint` clean; `bun run fmt` clean.

| test | why |
|---|---|
| uploads keyed by org | the archive lands under the org prefix, not a bare repo hash |
| skips without provenance | an archive with an unknown owner must not exist |
| steady-state sweep | recompresses nothing; mtime unchanged |
| restores on a cold root | the uploaded archive is actually readable by #5824's restore |
| two orgs, one repo | separate archives; neither overwrites the other |
| in-flight publish | a golden dir with no `node_modules` is skipped, so no partial tree ships |
| `.vite` excluded | pod-local churn does not travel |
| both no-op paths | missing cache root, missing remote root |

Inverted the `build-config-payload` test that asserted an operator-only payload, since the org is now sent too. Fixed `splitGoldenLockDir` to return empty strings alongside `ok=false` — a test caught it handing back a half-parsed key.

Chart render verified across: uploader off (nothing), on (DaemonSet + writable PV/PVC, distinct from the tenant's), dynamic path without `pvcName` (fails with the reason), dynamic path with it (renders, no PV).

**Not validated on a real pod.** Per the repo norm, "done" is a tailed pod showing an L2 hit; that needs both PRs deployed and `goldenUploader.enabled` on, which is a `deco-apps-cd` change after these land.

---
## Summary by cubic
Adds a node-level `golden-uploader` DaemonSet that moves dependency goldens from each node’s L1 cache to a shared L2 store. Keys are per‑org to prevent cross‑org restores; `orgId` now reaches the daemon and is written beside each golden. Helm chart bumped to 0.13.0 (uploader off by default).

- **New Features**
  - Node-scoped uploader (same image, `golden-uploader` subcommand) reads a read‑only hostPath and writes to a writable shared store; sequential sweeps with low CPU limits.
  - Shared archive paths: org → repo → `<pm>-<lock>.tar.zst`; existing archives are skipped; remote restore/publish require `orgId`.
  - Daemon writes `.golden-meta.json` (`orgId`/`cloneUrl`/`pm`); uploader skips goldens without a readable org.
  - Helm 0.13.0: DaemonSet plus a writable PV/PVC and validations; placement matches sandbox pods; default disabled.

- **Migration**
  - Enable with `goldenUploader.enabled: true` and ensure `depsCache.enabled` and `depsCache.remote.enabled` are true.
  - Provide a writable L2 view via `depsCache.remote.volume.attributes` (chart‑provisioned PV/PVC) or set `goldenUploader.pvcName` to an existing claim.
  - If using a dynamic `depsCache.remote.storageClassName`, you must set `goldenUploader.pvcName` (the StorageClass cannot express the read‑only/read‑write split).

<sup>Written for commit ce884146ddb60bdd018412190323a2378eebf839. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5843?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/deploy/helm/sandbox-env/Chart.yaml
+++ b/deploy/helm/sandbox-env/Chart.yaml
@@ -9,6 +9,20 @@ description: |
   releases coexist in the shared `agent-sandbox-system` namespace.
   Requires the sandbox-operator chart to already be installed.
 type: application
+# 0.13.0: opt-in `goldenUploader` (default off) — a DaemonSet that carries L1
+# goldens from a node's hostPath into the shared L2 store, plus its own WRITABLE
+# PV/PVC over that store. Without it the shared store stays empty: a sandbox
+# pod's mount is read-only by design, so nothing a tenant boots can publish
+# there. A DaemonSet rather than a sidecar so the sandbox PodSpec is untouched —
+# no per-pod request, no effect on packing density, nothing new that can degrade
+# the health probe — and so the writable mount lives once per node in
+# infrastructure instead of inside every tenant's pod. Archives are keyed by org,
+# which is what makes fleet-wide publishing acceptable: a repo hash comes from
+# the clone URL, so two orgs cloning one public template would otherwise share a
+# key. Requires depsCache.enabled + depsCache.remote.enabled and a writable
+# volume source (the new PV, or goldenUploader.pvcName — mandatory on the dynamic
+# storageClassName path, which cannot express the read-only/read-write split).
+# Additive; no change unless enabled.
 # 0.12.1: pin `updateStrategy.type: OnReplenish` on every SandboxWarmPool
 # (generic + tenant). It is already the operator's default as of v0.4.5, so
 # this renders no behaviour change today — it exists so an operator upgrade
@@ -92,7 +106,7 @@ type: application
 # 0.8.0: org-fs sidecar (orgFs.*) — privileged rclone mounter + main-container
 # mounts. NOTE: the publish workflow skips existing versions, so this MUST
 # stay ahead of the latest published chart (0.7.5 at time of writing).
-version: 0.12.11
+version: 0.13.0
 # appVersion tracks the studio-sandbox image version (image.tag default).
 appVersion: "1.45.0"
 kubeVersion: ">=1.30.0-0"
--- a/deploy/helm/sandbox-env/templates/_helpers.tpl
+++ b/deploy/helm/sandbox-env/templates/_helpers.tpl
@@ -363,6 +363,77 @@ agent-sandbox-system with every other release.
 {{- end }}
 
 {{/*
+Node-level uploader: DaemonSet, plus its own writable PV/PVC over the same
+store. A distinct name from goldenRemoteName because the two volumes differ in
+exactly the thing that matters — the tenant's is read-only at the mount, this
+one is not.
+*/}}
+{{- define "sandbox-env.goldenUploaderName" -}}
+{{- printf "studio-sandbox-golden-uploader-%s" (include "sandbox-env.envName" .) -}}
+{{- end }}
+
+{{/*
+Claim the uploader writes through. An explicit goldenUploader.pvcName wins and
+this chart renders no PV/PVC — the escape hatch for a store this chart does not
+manage.
+*/}}
+{{- define "sandbox-env.goldenUploaderClaimName" -}}
+{{- if .Values.goldenUploader.pvcName -}}
+{{- .Values.goldenUploader.pvcName -}}
+{{- else -}}
+{{- include "sandbox-env.goldenUploaderName" . -}}
+{{- end -}}
+{{- end }}
+
+{{- define "sandbox-env.goldenUploaderSelectorLabels" -}}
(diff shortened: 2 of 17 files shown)


pedrofrxncx (approved):

nicacioliveira:
Rebased on current `main` (both branches were **86 commits** behind) and added the piece that was missing to actually ship:

**`packages/sandbox/package.json` 1.44.6 → 1.45.0.** The `release-studio-sandbox` workflow reads that version as the image tag and *skips the entire build* when the tag already exists in GHCR (`1.44.6` is published). Merging without the bump would have published chart 0.13.0 while reusing the old image — so the DaemonSet would have shipped against a binary with **no `golden-uploader` subcommand** and exited on every sandbox node. Worse than not merging.

Also in the rebase:
- chart version **0.12.0 → 0.13.0**: `main` had already used 0.12.0 and 0.12.1 for tenantPools and the warm-pool `updateStrategy` pin, and is at 0.12.10 now.
- fixed my own shortcut: the "omits orgId when there is no tenant" test stubbed `repo` with `as never`, and `buildConfigPayload` has since started reading `repo.userName.trim()`. Replaced the cast with a real fixture — that cast is exactly what CLAUDE.md warns about, and the rebase is what exposed it.

Re-verified from scratch after the rebase, not from cache: `go build`, `go vet`, `go clean -testcache && go test ./internal/setup/` (26 pass), `helm lint`, `bun test build-config-payload` (7 pass), `bun run fmt` clean.

PR #5868, opened 2026-08-09
fix(oauth): let the native desktop app's local origin through GitHub import's OAuth flow

tlgimenes:
## Summary

The "Import from GitHub" flow in the desktop app failed end-to-end with `Load failed`, and once that was fixed, `redirect_uri is not allowed` — three separate gaps, none of which ever mattered for plain web, all specific to the desktop app's real-TLS `local.studio.decocms.com` origin:

- **`apps/native/.../upstream.rs` — `caller_origin`'s scheme fallback.** When the MCP SDK's protected-resource fetch carries no `Origin` header, the fallback hardcoded `http://`. The desktop app terminates real HTTPS on `local.studio.decocms.com`, so the SDK's own resource-URL check rejected the scheme mismatch outright.
- **Same file — OAuth metadata localization.** Once the scheme was fixed, discovery still failed cross-origin: `authorization_servers` / `token_endpoint` / `registration_endpoint` in the OAuth metadata point straight at the deployment, which the webview has to `fetch` itself, and the deployment's CORS policy doesn't allow that origin. Generalized the existing `resource`-field rewrite into `rewrite_origin_urls`, routing those through local-api's own already-working `/oauth-proxy/*` forwarding. **Deliberately excludes `authorization_endpoint`** — verified live that localizing it 401s GitHub's real consent page, since that's a top-level browser navigation to the deployment's own consent screen (never a fetch) and needs the deployment's own session cookie, not this proxy's.
- **`apps/api/src/api/app.ts` — CORS + redirect_uri allowlists.** Both `resolveCorsOrigin` and the `/authorize` proxy's `redirect_uri` check special-cased bare `localhost` but never the desktop app's real local origin, so even a correctly-shaped local-origin request still failed once it reached the deployment.

## Testing

- `bun test apps/api/src/api/app.test.ts` — 17 pass (new cases for both allowlist fixes, including lookalike-origin rejection: `local.studio.decocms.com.evil.example`, `evil.local.studio.decocms.com`)
- `cargo test -p local-api --lib` — 806 pass (new/updated cases for `caller_origin`'s scheme fallback and the RFC 9728/8414 metadata rewrite, including the `authorization_endpoint` exclusion)
- `cargo fmt --check` / `cargo clippy` / `bun run fmt` / `oxlint` — clean
- **End-to-end**: ran `apps/api` locally with this fix applied, pointed a `DECOCMS_UPSTREAM_URL`-overridden desktop build at it, and completed a real GitHub OAuth grant + repo import against the real, already-deployed `github-mcp.decocms.com` service — confirmed a working repo-scoped connection (real installation ID, permissions, repository ID) was created.

Note: the `apps/api` fix requires a deploy to `studio.decocms.com` to take effect against the production desktop app — the native-side fixes are self-contained and ship with the next app build.

---
## Summary by cubic
Fixes the desktop app’s “Import from GitHub” OAuth by accepting the app’s local TLS origin and localizing discovery URLs to avoid CORS, so the flow completes end-to-end. Also tightens allowlists and corrects scheme detection in the native proxy.

- **Bug Fixes**
  - Native proxy: use the listener scheme for `Origin` fallback (no more hardcoded `http`), so MCP resource checks pass on `https://local.studio.decocms.com:*`.
  - Native proxy: localize OAuth discovery docs to the caller origin for both protected-resource and auth-server metadata; proxy `authorization_servers`, `token_endpoint`, and `registration_endpoint`; intentionally do not localize `authorization_endpoint`.
  - API: allow `https://local.studio.decocms.com:*` in CORS reflection and in `redirect_uri` validation via `isAllowedOAuthRedirectUri`; reject lookalike hosts.

- **Migration**
  - Deploy `apps/api` to apply CORS/redirect changes in production; native fixes ship with the next desktop app build.

<sup>Written for commit 0f47153454bb5bd17474260c34c91e89bd33b70b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5868?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/api/app.ts
+++ b/apps/api/src/api/app.ts
@@ -313,10 +313,24 @@ export function isSsoExemptPath(path: string): boolean {
  * cookie-authenticated cross-site request against this API and read the
  * response — a permissive-CORS-with-credentials hole. Only reflect
  * localhost/127.0.0.1 (the Vite dev server, on a different port than the
- * API) and the deployment's own origin, falling back to the request's own
- * origin when `baseUrl` isn't configured (same fallback already used for
- * redirect_uri validation above).
+ * API), the native desktop app's local control origin (see below), and the
+ * deployment's own origin, falling back to the request's own origin when
+ * `baseUrl` isn't configured (same fallback already used for redirect_uri
+ * validation above).
  */
+/**
+ * The native desktop app terminates local TLS on its own real domain
+ * (`local.studio.decocms.com`, any port — see apps/native's
+ * `control_origin.rs`/`local_tls.rs`) rather than `localhost`, because a
+ * secure context needs a real domain and per-sandbox cookie isolation needs
+ * one the browser groups as a single site. Exact hostname match only — a
+ * subdomain or suffix (`evil.local.studio.decocms.com`,
+ * `local.studio.decocms.com.evil.example`) must NOT pass.
+ */
+function isNativeDesktopControlOrigin(hostname: string): boolean {
+  return hostname === "local.studio.decocms.com";
+}
+
 export function resolveCorsOrigin(
   origin: string,
   {
@@ -333,6 +347,16 @@ export function resolveCorsOrigin(
   if (originHost === "localhost" || originHost === "127.0.0.1") {
     return origin;
   }
+  // A connection's OAuth `authorization_servers` / `token_endpoint` /
+  // `registration_endpoint` legitimately point straight at this deployment
+  // (oauth-proxy.ts deliberately doesn't proxy those — see
+  // `apps/native/crates/local-api/src/routes/upstream.rs`'s
+  // `rewrite_origin_urls`), so the desktop webview calls them cross-origin —
+  // without this, every discovery/register/token fetch in that flow fails
+  // with no ACAO header ("Load failed" client-side).
+  if (isNativeDesktopControlOrigin(originHost)) {
+    return origin;
+  }
 
   const allowedOrigin = baseUrl ?? requestOrigin;
   try {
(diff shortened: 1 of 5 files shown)


apps/api/src/api/app.ts
```diff
@@ -333,6 +347,16 @@ export function resolveCorsOrigin(
...
     return origin;
   }
+  // A connection's OAuth `authorization_servers` / `token_endpoint` /
+  // `registration_endpoint` legitimately point straight at this deployment
+  // (oauth-proxy.ts deliberately doesn't proxy those — see
+  // `apps/native/crates/local-api/src/routes/upstream.rs`'s
+  // `rewrite_origin_urls`), so the desktop webview calls them cross-origin —
+  // without this, every discovery/register/token fetch in that flow fails
+  // with no ACAO header ("Load failed" client-side).
+  if (isNativeDesktopControlOrigin(originHost)) {
```
tlgimenes, line 357: maybe coallesce with the if on line 347 to avoid lots of branching in code

tlgimenes: Disagree — keeping these separate. The two branches have unrelated justifications: `localhost`/`127.0.0.1` is "any local dev tooling on loopback" (no comment needed), while `isNativeDesktopControlOrigin` needs the 6-line explanation of why this specific production domain is allowed cross-origin (the `authorization_servers`/`token_endpoint`/`registration_endpoint` proxy-bypass rationale). Coalescing into one condition would force that comment to either sit above the combined `if` — misleadingly implying it also explains the `localhost` case — or stay separated from the check it describes. In a `credentials: true` CORS allowlist, comment-to-condition locality matters more than branch count. For contrast, `isAllowedOAuthRedirectUri` below does combine three conditions into one expression, but each disjunct there needs at most a clause of explanation, not a standalone rationale — different shape from this case.

PR #5919, opened 2026-08-11
fix(sandbox): respawn dead dev server instead of endless "starting"

guitavano:
## Summary

Fixes the recurring **"Blocos indisponíveis" / "Server is starting…"** dead end where a sandbox preview never recovers after the tab goes idle — and clicking **Retry** doesn't help.

Diagnosed from a user HAR: after ~10min idle the pod's dev server dies, but nothing respawns it. The in-pod proxy serves `StartingHTML` (503) forever (a connection error to a **known-but-dead** port), Studio's Blocks tab shows `errorStateSandbox`, and the web Retry fires a `SANDBOX_START` that returns `isNewVm:false` — reusing the wedged VM, so it's a no-op. The control plane is healthy throughout; the failure is entirely in the sandbox data plane.

## What changed (daemon, Go)

**1. Dev-server liveness watchdog**
- `internal/devwatch/devwatch.go` (new) — pure `Decide()` → restart / give-up / none, unit-tested.
- `main.go` `devWatchLoop` — when the probe reports not-serving in a restartable phase (running/starting/crashed) with a **known** port (sniffed OR previously-served — a *merely-configured* `application.port` does **not** count, so a slow first boot is never interrupted) for `GracePeriod` (default 20s), it resets sniffer+prober and calls `orchestrator.RestartDev()`. Bounded by `MaxRestarts` (3); on exhaustion it transitions to a real `start-failed` (honest error instead of an endless "Server is starting…").

**2. Un-latch the crash `error` status so restarts aren't skipped**
- `orchestrator.RestartDev()` clears the `error` a non-zero dev-script exit latches (else `stepStartInner`'s `status != running` guard skips every start), force-stops the dev task, and re-enqueues `StepStart`.
- `Handle()` clears the same latch on `bootstrap`/`branch-change`, so a `SANDBOX_START` reclaim onto a VM with a crashed dev server actually restarts it.

**Tunables (defaults unchanged):** probe cadence `SANDBOX_PROBE_FAST_MS`/`_SLOW_MS`, watchdog windows `SANDBOX_DEV_WATCH_GRACE_MS`/`_TICK_MS`. The 30s healthy-probe `SlowInterval` bounds detection latency and is now adjustable.

## User impact

Before: idle → dead preview → "Blocos indisponíveis" forever, Retry no-op. After: the daemon respawns the dev server automatically (no click); if it genuinely can't stay up after 3 tries, it shows a real error instead of looping.

Note: the `[::1]` proxy dial is **not** a cause — the proxy transport already uses `probe.DialLoopback`, which falls back to `127.0.0.1`.

## Testing

- `go test ./...` (all daemon packages), `go vet`, `gofmt` — pass
- `internal/devwatch` unit test (10 cases)
- **New black-box e2e** `daemon-e2e/daemon.devwatch.e2e.test.ts` — kills the dev server over the wire, asserts auto-recovery to `running` + the restart log; stable across runs (~5–6s)
- `daemon.proxy` + `daemon.warmpool` e2e — 15/15, no regression
- `tsc --noEmit` — pass

## Affected areas

`packages/sandbox/daemon-go` (probe, setup/orchestrator, main), plus one new e2e in `packages/sandbox/daemon-e2e`. No Studio server/web changes.

## Follow-up (not in this PR)

Wire the web **Retry** button to trigger a real daemon dev-server restart (today a `SANDBOX_START` on an unchanged config is a daemon no-op). With the automatic watchdog, the click is rarely needed.

---
## Summary by cubic
Automatically respawn dead dev servers to stop endless “Server is starting…” loops in sandbox previews. Handles slow first boots and flapping crashes correctly, and shows a real start-failed after repeated attempts.

- **Bug Fixes**
  - Add a liveness watchdog that restarts the dev server when not-serving in restartable phases and a port is known; grace now starts at “port known,” not boot. Reset sniffer/prober and call `RestartDev()`; cap attempts (default 3), then transition to start-failed.
  - Bound flapping: the restart budget resets only after continuous serving for a stable window, so brief serves can’t re-arm the budget indefinitely.
  - Unlatch dev crash errors so restarts aren’t skipped: `RestartDev()` force-stops the dev task and re-enqueues start; also clear the latch on bootstrap/branch-change so a `SANDBOX_START` reclaim actually restarts.
  - Make probe cadence and watchdog knobs tunable and safe: `SANDBOX_PROBE_FAST_MS`, `SANDBOX_PROBE_SLOW_MS`, `SANDBOX_DEV_WATCH_GRACE_MS`, `SANDBOX_DEV_WATCH_TICK_MS`, `SANDBOX_DEV_WATCH_STABLE_MS`, `SANDBOX_DEV_WATCH_MAX_RESTARTS`; clamp to sane floors and read liveness via `probe.Prober.Current()` to avoid stale mirrors.

<sup>Written for commit 7e8e591e63fc2c5e01378f489737d2cea4eec659. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/daemon-e2e/daemon.devwatch.e2e.test.ts
+++ b/packages/sandbox/daemon-e2e/daemon.devwatch.e2e.test.ts
@@ -1,0 +1,124 @@
+/**
+ * Daemon conformance suite — DEV-SERVER LIVENESS WATCHDOG.
+ *
+ * When the dev server dies out from under a live daemon (idle freeze, crash, a
+ * stale port after a reclaim), nothing used to respawn it: the proxy served
+ * "Server is starting…" forever and the preview never came back. The daemon now
+ * watches liveness and respawns a dead-but-known-port dev server, bounded, then
+ * surfaces a real start-failed if it can't stay up.
+ *
+ * Black-box: we drive a real `dev` HTTP server, kill it over the wire, and
+ * assert the daemon brings it back to `running` on its own. The watchdog's
+ * grace/tick are shrunk via env so the test doesn't wait the 20s default.
+ */
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+
+import {
+  authHeaders,
+  type BareRepo,
+  bootstrapRepo,
+  type Daemon,
+  HOOK_TIMEOUT_MS,
+  freePort,
+  readSseUntil,
+  setupBareRepo,
+  startDaemon,
+  stopDaemon,
+  url,
+  waitForOrchestratorIdle,
+} from "./daemon.e2e.helpers";
+
+const SETUP_TIMEOUT_MS = 60_000;
+
+describe("daemon e2e: dev-server liveness watchdog", () => {
+  let d: Daemon;
+  let repo: BareRepo;
+  let devPort: number;
+
+  beforeEach(async () => {
+    devPort = await freePort();
+    // A dev server that actually serves HTML, so the daemon reaches `running`.
+    repo = setupBareRepo({
+      withPackageJson: true,
+      scripts: {
+        dev: `node -e "require('http').createServer((_q,s)=>{s.setHeader('content-type','text/html');s.end('<html>ok</html>')}).listen(process.env.PORT||3000)"`,
+      },
+    });
+    // Shrink the probe cadence and watchdog windows so a dead dev server is
+    // detected and respawned in a couple of seconds instead of the 30s slow
+    // probe + 20s production grace.
+    d = await startDaemon({
+      SANDBOX_PROBE_FAST_MS: "300",
+      SANDBOX_PROBE_SLOW_MS: "1000",
+      SANDBOX_DEV_WATCH_GRACE_MS: "1000",
+      SANDBOX_DEV_WATCH_TICK_MS: "500",
+    });
+  }, HOOK_TIMEOUT_MS);
+
+  afterEach(async () => {
+    await stopDaemon(d);
+    repo.cleanup();
+  }, HOOK_TIMEOUT_MS);
+
+  const lifecyclePhase = async (): Promise<string> => {
+    const { text } = await readSseUntil(url(d, "/_sandbox/events"), {
+      headers: authHeaders(),
+      predicate: (acc) => acc.includes("event: lifecycle"),
+    });
+    return (
+      (
+        JSON.parse(
+          /event: lifecycle\ndata: (.*)\n/.exec(text)?.[1] ?? "{}",
+        ) as { state?: { phase?: string } }
+      ).state?.phase ?? ""
+    );
+  };
+
+  const waitForPhase = async (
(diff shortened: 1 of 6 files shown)


guitavano:
## Addressed a multi-perspective review (2nd commit)

Ran a parallel critique (correctness / testing / architecture / performance / scope). Two independent critics found the same correctness blocker; fixed it plus the important follow-ups in `7e8e591`.

**🔴 Blocker — grace bypassed on first boot.** `notServingSince` started on the first not-serving tick (during *install*) and accumulated across the whole boot, so a newly-`starting` server was restarted with **zero grace** → slow-compiling apps (Next/faststore) would be killed and marched to a spurious `start-failed`. The instant-serving test masked it. Fixed by moving the clock/budget into a unit-testable `devwatch.Tracker` (injected clock) where the clock runs **only while a restart could help** (restartable phase + known port + not serving) — grace is now measured from *port-known*, per the field's contract.

**🟠 Flapping budget.** The restart budget reset on a single serving tick → a crash-loop that briefly serves could re-arm forever, defeating the bound. Now resets only after **continuous serving ≥ StableWindow (60s)**, so `MaxRestarts → start-failed` actually holds.

**🟠 Testing gaps.**
- New e2e proving a **branch-change after a non-zero dev crash** (`start-failed` + latched error) recovers to `running` — exercises `clearCrashError` (fix #2), which the intentional-kill case structurally couldn't.
- `devwatch.Tracker` unit table: grace-from-port-known (blocker regression), no-restart-without-port, give-up after MaxRestarts, flapping-does-not-reset-budget, sustained-serving-resets, phase-leave disarm.

**Minor.** Consume `probe.Prober.Current()` synchronously instead of an async `probeOnline` mirror; clamp env intervals to a floor (no hot-spin); expose `SANDBOX_DEV_WATCH_STABLE_MS`/`_MAX_RESTARTS`; softened the `DefaultGracePeriod` doc (heuristic, not guarantee); confirmed the `[::1]` proxy dial is **not** a bug (transport already falls back to `127.0.0.1` via `probe.DialLoopback`).

### Deferred (follow-ups, non-blocking)
- **Architecture:** the watchdog goroutine still lives in `main.go`; a full `devwatch.Watcher` component owning the loop (and `RestartDev` owning the sniffer/prober reset) is a reasonable next step. The fragile decision state is now in the tested package, which was the core of the critique.
- **Minor:** `start-failed` self-recovery when a wedged process later answers on its own; give-up counting a no-op `stepStart` skip; quiescing the watchdog before the shutdown publish.

PR #5966, opened 2026-08-12
fix(sandbox): a pod that dies mid-run is infrastructure, not a cancel

pedrofrxncx:
Found debugging `board_46FLuBSK-bxaqPQ3_t4E1` (deco-studio): 6 runs, all failed, no PR, `run_count` past the cap.

## 1. `cancelled: run cancelled` is still happening — #5859 only fixed half

Three of that card's six runs died with it, on 08-11, four days after #5859. That PR handled the **displaced** case (takeover → `superseded`). The other branch is unchanged:

```go
// Genuinely cancelled: the client hung up, or a DELETE stopped the run…
body.write(terminalFrame("cancelled", "run cancelled"))
```

Those two causes are not the same thing. On SIGTERM the daemon calls `dispatchReg.CancelAll()`, every in-flight dispatch lands here, and Studio reads `cancelled` as the run's verdict — `DELIBERATE_KINDS`, zero retries. So **a sandbox pod being evicted or scaled in permanently kills the turn**, and it does it on the most recoverable failure there is: `shutdown()` runs `gitx.Publish` immediately after, so the work is already committed and pushed to the branch.

The discriminator was already there and unused — `HandleCancel` writes `reg.tombstones[runId]`, which is the only proof a stop was *asked for*. Now:

| | terminal | Studio |
|---|---|---|
| displaced by a takeover | `superseded` | stop quietly, successor owns it |
| tombstoned (a DELETE) | `cancelled` | fail, no retry — a human said stop |
| **neither** | **`sandbox_gone`** | **`SandboxUnreachableError` → continue on a fresh pod** |

## 2. No `SandboxUnreachableError` message was recognized as infrastructure

There are five of them (`the sandbox stream broke mid-run: …`, `no output from the sandbox for 90s`, `the sandbox stopped streaming after N chunks`, …). `TRANSIENT_ERROR_PATTERNS` matched **none**, so the failure the board sees most often got `UNKNOWN_RETRY_BUDGET = 1` instead of `TRANSIENT_RETRY_BUDGET = 3`. Every "retry 1 of 1" on that card's timeline is this.

Fixed by marker, not by wording: the constructor stamps `[SANDBOX_UNREACHABLE]` (same convention as `[SUBSCRIPTION_REQUIRED]` / `[CREDITS]`) and `isTransientRunFailure` checks for it. A reworded message stays covered; so does a throw site written next year.

## 3. The forwarder's error branch was silent

`ws.on("error")` invalidated the record and cleaned up without logging. That is the branch that fires when an *established* port-forward dies mid-run — i.e. the source of "The socket connection was closed unexpectedly" — so **why** it keeps happening has no answer anywhere in the logs. It now logs the cause chain like the sibling `.catch` already does. I still can't tell you the root cause of the drops; this is what makes the next one answerable.

## Testing

- Go: `TestTombstoneMarksOnlyAskedForCancels` (shutdown/dropped ≠ DELETE, and the tombstone is per-run), `TestUnauthorizedCancelLeavesNoTombstone` (a rejected DELETE must not turn a continuable failure into a permanent one), `sandbox_gone` added to the terminal-frame table. `go test ./internal/dispatch/` ok.
- TS: `errorForTerminal` extracted pure from the stream loop (it replaces three inline branches) and unit-tested for all three outcomes; the five real `SandboxUnreachableError` messages asserted transient **through the real constructor**, so dropping the marker fails the test; a plain error mentioning "sandbox" asserted *not* transient.
- `bun test` on both touched files: 43 pass. `bun run check`, `lint`, `fmt` clean.

## Rollout

The daemon ships in the sandbox image, Studio in the API — they roll separately, and either order is safe: an old Studio sees `sandbox_gone` and fails the run exactly as it fails `cancelled` today (no worse), and an old daemon keeps sending `cancelled`, which a new Studio still treats as deliberate. The improvement lands when both are out.

---
## Summary by cubic
Treat sandbox pods that die mid-run as infrastructure, not user cancels. Adds `sandbox_gone` → `SandboxUnreachableError`, renews the sandbox claim TTL while streaming to prevent 15‑minute evictions, continues runs on a fresh pod with the correct retry budget, and logs dropped port-forward connections.

- **Bug Fixes**
  - Daemon: emit `sandbox_gone` when a run ends without a tombstone (pod shutdown/connection drop). Keep `cancelled` only for real DELETEs. Studio maps `sandbox_gone` to `SandboxUnreachableError` and continues the turn.
  - Dispatch client: renew claim TTL during streaming (immediately and every 5 minutes) and stop on exit, so headless runs don’t hit the 15‑minute shutdown.
  - Task board: mark all `SandboxUnreachableError` messages with `SANDBOX_UNREACHABLE_PREFIX` and treat them as transient (full retry budget).
  - Agent sandbox: log cause chain on `ws.on("error")` for established port-forward drops.

- **Migration**
  - Backward compatible. Old Studio treats `sandbox_gone` like `cancelled`; new Studio still handles old `cancelled`. Improvement appears when both daemon and Studio are deployed.

<sup>Written for commit 7bf1713ffc028a57ce467a90e390e9ba75132304. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/5966?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/harnesses/sandbox-dispatch-client.test.ts
+++ b/apps/api/src/harnesses/sandbox-dispatch-client.test.ts
@@ -3,6 +3,7 @@ import type { UIMessageChunk } from "ai";
 import { harnessRunResultSchema } from "@decocms/sandbox/dispatch/schemas";
 import {
   dispatchWithContinuation,
+  errorForTerminal,
   harnessRunsInSandbox,
   isRunSuperseded,
   isUnreachableStatus,
@@ -308,6 +309,35 @@ describe("dispatchWithContinuation", () => {
     });
     await expect(collect(run)).rejects.toThrow("a newer dispatch took over");
     expect(attempts).toBe(1);
+  });
+});
+
+// Which terminal code continues the turn, which drops it quietly, and which
+// fails it. `sandbox_gone` is the one that changed: the daemon reported an
+// evicted pod as `cancelled`, so a turn nobody stopped was settled as a
+// deliberate cancel and never retried.
+describe("errorForTerminal", () => {
+  test("a sandbox that could not finish is continuable, not a failure", () => {
+    const err = errorForTerminal("sandbox_gone", "the sandbox stopped mid-run");
+    expect(err).toBeInstanceOf(SandboxUnreachableError);
+    expect(isRunSuperseded(err)).toBe(false);
+  });
+
+  test("a takeover stops this attempt quietly", () => {
+    const err = errorForTerminal("superseded", "a newer dispatch took over");
+    expect(isRunSuperseded(err)).toBe(true);
+    expect(err).not.toBeInstanceOf(SandboxUnreachableError);
+  });
+
+  // A cancel a human asked for, and a harness that really crashed, are the
+  // run's own outcome — continuing either would re-run work that already ended.
+  test("a real terminal fails the run", () => {
+    for (const code of ["cancelled", "harness_crashed", "bad_input"]) {
+      const err = errorForTerminal(code, "why");
+      expect(err).not.toBeInstanceOf(SandboxUnreachableError);
+      expect(isRunSuperseded(err)).toBe(false);
+      expect(err.message).toBe(`${code}: why`);
+    }
   });
 });
 
--- a/apps/api/src/harnesses/sandbox-dispatch-client.ts
+++ b/apps/api/src/harnesses/sandbox-dispatch-client.ts
@@ -485,6 +485,42 @@ function toWireInput(input: HarnessStreamInput): unknown {
   return wire;
 }
 
+/**
+ * How often a streaming dispatch pushes its sandbox's shutdown deadline out.
+ * Same cadence as the preview SSE handler's `TTL_RENEW_MS`, and for the same
+ * reason: comfortably inside the 15-minute claim TTL, so a missed renewal (or a
+ * slow apiserver) still leaves two more chances before the pod is reaped.
+ */
+const TTL_RENEW_MS = 5 * 60_000;
+
+/**
+ * Keep the claim alive on an interval, returning the stop. Best-effort by
+ * construction — a failed renewal costs the run its pod at the deadline, which
+ * is exactly today's behavior, so it must never break the stream it is
+ * protecting.
+ */
+function renewWhileStreaming(
+  provider: SandboxProvider,
+  handle: string,
+): () => void {
+  if (!provider.renewTtl) return () => {};
+  const renew = () =>
+    void provider
+      .renewTtl?.(handle)
+      .catch((err) =>
+        console.warn(
+          `[sandbox-dispatch] TTL renew failed for ${handle}: ${err instanceof Error ? err.message : String(err)}`,
+        ),
+      );
+  // Immediately as well as on the interval: a run dispatched onto a reused
+  // sandbox is adopting a TTL that is already part-spent — that is how a run
+  // could die four minutes in.
+  renew();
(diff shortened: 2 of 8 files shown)


pedrofrxncx:
## Root cause found — it's the claim TTL, not the housekeeper. Fix pushed to this branch.

**A SandboxClaim is born with a 15-minute death sentence that only a browser can postpone.**

`runner.ts` creates every claim with `spec.lifecycle.shutdownTime = now + DEFAULT_IDLE_TTL_MS` (15 min) and the operator deletes the pod at that instant. `renewTtl()` pushes the deadline out — and it had exactly **one** caller:

```
apps/api/src/api/routes/sandbox-events-handler.ts:124:  const renew = () => void runner.renewTtl?.(claimName);
```

That's the preview SSE stream. A headless Super Agent run has no browser attached, so nothing ever renewed it. Every `claude-code` run had a hard 15-minute wall clock; at the deadline the operator SIGTERMs the daemon, `CancelAll()` fires, and the turn dies mid-edit — which is precisely the `cancelled` → `sandbox_gone` case this PR started with.

Confirmed live in prod (`kubectl get sandboxclaims -n agent-sandbox-system`):

```
NAME                                   AGE                    SHUTDOWN                   Δ
conn-9cwwg7ngwnqyqm6r4ek-9ac885e0…     2026-08-12T13:13:58Z   2026-08-12T13:28:58Z       15:00
miguel-duqx2qsm-8a29bceea4b11ad4       2026-08-12T13:13:21Z   2026-08-12T13:28:21Z       15:00
cmarinho-y58w2qsm-49897b6939d10bb5     2026-08-12T12:41:39Z   2026-08-12T13:36:40.182Z   55:01  ← SSE attached
conn-y-drojqof6pk0qfm99a-99d4d1dd…     2026-08-12T12:45:38Z   2026-08-12T13:35:38.910Z   50:00  ← SSE attached
```

Everything with a browser on it drifts to 30–55 minutes. Everything else is exactly `creation + 15:00`. The six runs on `board_46FLuBSK…` died 11–17 min after thread creation, which is 15 min from *claim* creation once queue wait is subtracted.

### Why it is NOT the housekeeper

The other suspect, and it's clean:

- it reaps on the daemon's own `/idle`, and a streaming run bumps activity on every frame **plus** a 15s keepalive, so idle stays ~0;
- every `SandboxReaped` event in prod is a genuinely idle pod (`idle_ms=654499`, `632260`, `611048`, … against `ttl_ms=600000`);
- structurally it can only end a claim **early**, never extend one. The claim's deadline is a separate clock, and `renewTtl` is the only thing that moves it.

### The fix

`dispatchToDaemon` now renews on the same 5-minute cadence as the SSE handler, immediately on dispatch as well (a run landing on a **reused** sandbox adopts a TTL that is already part-spent — that's how a run could die four minutes in), and clears the interval in a `finally` so an abandoned generator can't leak one that keeps a dead claim alive.

This is the actual bug. The `sandbox_gone` relabelling and the retry-budget marker in the earlier commits are still worth having — they're what makes the *remaining* pod deaths (spot reclaim, node loss, real evictions) recoverable instead of terminal — but this is what makes a 20-minute task possible at all.

PR #6002, opened 2026-08-12
fix(analytics): instrument the report-funnel gaps found in the PostHog audit

rafavalls:
## Summary

Follow-up to a PostHog instrumentation audit of the diagnostic-report → kanban → payment funnel. Fixes two verified, previously-uncovered gaps:

- **Kanban push had zero instrumentation.** `task-board-import.ts` (the diagnostic-worker → task-board endpoint) had no `posthog.capture` calls at all. Adds `task_board_import_landed` (created/updated/skipped/delegated + `run_id`), fired once per non-deduped import, after commit. This is the **receiver's** view of the same moment the reports engine already emits as `diagnostic_tasks_pushed` (the **sender's** view, from `pushBacklogToStudio`) — the two carry the same `run_id` and are meant to be joined, not summed.
- **Report-side and onboarding-side events weren't linkable.** `report_cta_clicked` (report deck) and `commerce_onboarding_*` (onboarding flow) had no shared attribution. `onboardingUrl()` now carries the report's attribution (`entrypoint`, `share_id`, `email_run_id`, `utm_*`) through the URL, echoed back on `commerce_onboarding_viewed` and the companion connect/disconnect events. This composes with reports#341's server-side stitching (LP visitor's PostHog id persisted on the diagnostic) rather than overlapping it.

**Dropped after review**: the `board_paywall_checkout_*` instrumentation on `backlog-paywall.tsx`. Caught in review: `start_checkout` and the `locked` flag it depends on are both gone from the reports engine (reports#344), so the banner never renders and the tool call can't succeed — instrumenting it would have produced a chart that reads as "nobody clicks unlock" when the truth is "there is no button." The component itself is dead UI, being removed in #6005.

**Correction from review**: I'd originally noted the client-side funnel and server-side Stripe webhook events (`subscription_started` etc.) couldn't be joined due to different distinct_id schemes. That's wrong as of #5860 — `captureOrgEvent` sets `distinctId: "org:<id>"` **and** `groups: { organization: <id> }`, and the client-side events here carry `organization_id`. The join exists at the **organization/group level**, just not at the person level (a webhook has no acting user). Left as a follow-up, not included here.

## Test plan
- [x] `bun run check` — all workspaces pass, including `decocms` (api) and `@decocms/studio-web`
- [x] `bun run lint` — 0 errors; 0 warnings in any touched file
- [x] `bun run fmt` — no changes needed
- [x] `bun test` on the touched frontend areas (commerce-onboarding, reports) — 152 pass / 0 fail
- [x] `bun test` on `task-board-import` — pre-existing integration-test failures confirmed identical with/without this change (require a real Postgres not available in this sandbox); nothing newly broken
- [ ] Manual: trigger a diagnostic → kanban push and confirm `task_board_import_landed` shows up in PostHog, joinable to `diagnostic_tasks_pushed` on `run_id`

### DIFF
--- a/apps/api/src/api/routes/task-board-import.ts
+++ b/apps/api/src/api/routes/task-board-import.ts
@@ -389,9 +389,9 @@ export const createTaskBoardImportRoutes = () => {
       }
     }
 
-    // Funnel event: the only signal a diagnostic run reached the board.
+    // Receiver's counterpart to the engine's diagnostic_tasks_pushed — join on run_id, never sum.
     captureOrgEvent({
-      event: "report_tasks_pushed",
+      event: "task_board_import_landed",
       organizationId,
       properties: {
         created: outcome.created,
--- a/apps/api/src/api/routes/task-board-import.ts
+++ b/apps/api/src/api/routes/task-board-import.ts
@@ -6,6 +6,7 @@ import {
   nextTagColor,
   SUPER_AGENT_ASSIGNEE_ID,
 } from "@decocms/shared/task-board";
+import { captureOrgEvent } from "@/posthog";
 import { TagStorage } from "@/storage/tags";
 import { TaskBoardStorage } from "@/storage/task-board";
 import type { TaskBoardItem } from "@/storage/types";
@@ -387,6 +388,23 @@ export const createTaskBoardImportRoutes = () => {
           });
       }
     }
+
+    // Funnel event: the only signal a diagnostic run reached the board.
+    captureOrgEvent({
+      event: "report_tasks_pushed",
+      organizationId,
+      properties: {
+        created: outcome.created,
+        updated: outcome.updated,
+        skipped: outcome.skipped,
+        delegated,
+        quota_blocked: quotaBlocked,
+        ...(runId ? { run_id: runId } : {}),
+        ...(parsed.data.source?.url
+          ? { source_url: parsed.data.source.url }
+          : {}),
+      },
+    });
 
     return c.json({
       created: outcome.created,
--- a/apps/web/src/layouts/task-board/backlog-paywall.tsx
+++ b/apps/web/src/layouts/task-board/backlog-paywall.tsx
@@ -88,11 +88,13 @@ function BacklogPaywallModal({
   open,
   onOpenChange,
   cdClient,
+  organizationId,
   onExpired,
 }: {
   open: boolean;
   onOpenChange: (open: boolean) => void;
   cdClient: CommerceDiscoveryClient | null;
+  organizationId: string;
   /** Called when checkout can't start because the report app is the only place
    *  that can (no client) — falls back to opening it. */
   onExpired: () => void;
@@ -101,7 +103,14 @@ function BacklogPaywallModal({
   const [starting, setStarting] = useState(false);
 
   const startCheckout = async () => {
+    track("board_paywall_checkout_started", {
+      organization_id: organizationId,
+    });
     if (!cdClient) {
+      track("board_paywall_checkout_failed", {
+        organization_id: organizationId,
+        reason: "no_client",
+      });
       onExpired();
       return;
     }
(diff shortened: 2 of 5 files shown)


viktormarinho (commented):
Reviewed against `main` as of today. Two of the three parts are good and land in gaps nothing else covers. The paywall part instruments a flow that no longer exists — details below.

## ✅ `report_tasks_pushed` (`apps/api/src/api/routes/task-board-import.ts`)

Good, and genuinely complementary rather than duplicated. The engine side already emits `diagnostic_tasks_pushed` (reports#327/#339) from `pushBacklogToStudio`, but that's the **sender's** view: what it attempted, plus `pushed`/`deduped`/`reason`/`pick_error`. Yours is the **receiver's** view, and it carries two things the engine can never see: `skipped` and `quota_blocked`. Both events carry `run_id`, so they join cleanly. Placement is right too — after commit, and outside the dedupe early-return, so a replayed import doesn't double-count.

Two small asks:

1. **Say in the description that these are a pair.** Two events for one logical moment is a foot-gun for whoever builds an insight later — someone will sum them. A line like "the engine-side counterpart is `diagnostic_tasks_pushed`; join on `run_id`, never add them" in both descriptions kills that.
2. **Naming.** We just settled a taxonomy in reports#339: `finding` → `actions` (LLM proposals, every run) → `tasks` (board cards, born at push) → `runs` (auto-fix). Engine-emitted events are `diagnostic_*`. `report_tasks_pushed` reads like the engine event but isn't — consider `task_board_import_landed` (or anything that names the receiver), so the two are distinguishable at a glance in the events list.

## ❌ `board_paywall_checkout_*` (`apps/web/src/layouts/task-board/backlog-paywall.tsx`) — the flow is dead

This is the one I'd hold. Three independent facts, each verifiable on `main`:

1. **The tool being called doesn't exist.** `startCheckout` calls `cdClient.callTool({ name: "start_checkout" })`. In the reports engine, `start_checkout` is gone — `git grep start_checkout -- api/` in `decocms/reports@main` returns nothing. It was dropped before the checkout cleanup, and reports#344 (merged yesterday) removed the remaining machinery: `stripe.ts`, `/checkout/success`, `/checkout/cancel`, `/stripe/webhook`, the `entitlement` table.
2. **The banner never renders.** `BacklogPaywallBannerInner` bails on `if (!diagnostic?.locked) return null`, and `locked` is no longer a field on the engine's `publicView`/`ownerView` payload. It's always `undefined`.
3. **Therefore the events can't fire as intended.** `board_paywall_checkout_opened` is unreachable by construction; the other two would only fire from a component that doesn't mount.

The risk isn't just "an empty chart" — it's an empty chart that *reads* as "nobody clicks unlock" when the truth is "there is no button." That's worse than no instrumentation.

Suggested change: drop this part from the PR. The component itself is now dead UI and should go — I'm opening a follow-up PR to delete it. If we want a kanban paywall again, it's a different flow now: monetization is the **$50/mo org subscription**, whose checkout is `ORGANIZATION_BILLING_CHECKOUT_START`, already instrumented with `subscription_checkout_started` (#5918).

## ✅ Report → onboarding attribution (`reports/onboarding.ts`, `commerce-onboarding.tsx`, `use-connect-companion.ts`)

Useful, and it closes a join nobody had. It also composes with the server-side stitching in reports#341 (the LP visitor's PostHog id is now persisted on the diagnostic and reused on `diagnostic_run_completed`/`_claimed`) rather than overlapping with it.

## One correction to the "not included" note

> joining the client-side funnel to the server-side Stripe webhook events (`subscription_started` etc.), which use a different distinct_id scheme entirely

That changed with #5860. The subscription events are emitted through `captureOrgEvent`, which sets `distinctId: "org:<id>"` **and** `groups: { organization: <id> }`, with `$process_person_profile: false` so orgs don't pollute the persons table. The client-side events here carry `organization_id`. So the join does exist — it's **at the organization level** (group analytics), not person level. Person-level never will work across that boundary, because a webhook has no acting user. Worth doing, since it's the analysis you set aside.

viktormarinho:
Follow-up on the paywall point: opened #6005, which deletes `backlog-paywall.tsx` (banner + modal + its i18n strings + the `R$ 99 / R$ 499` offer copy). If that lands, the `board_paywall_checkout_*` block here has nothing left to instrument — the other two parts of this PR stand on their own and I'd merge them as-is (modulo the naming/pairing notes above).

rafavalls:
Addressed in fa580bb7a:

1. **Dropped `board_paywall_checkout_*` entirely** and reverted `backlog-paywall.tsx` to its `main` state. Confirmed `start_checkout` and `locked` are gone from the reports engine as described — agreed instrumenting an unreachable flow is worse than not instrumenting it. Will let #6005 take the component out.
2. **Renamed `report_tasks_pushed` → `task_board_import_landed`**, with a one-line comment noting it's the receiver's counterpart to the engine's `diagnostic_tasks_pushed` and that they join on `run_id` rather than sum.
3. **Updated the PR description's incorrect claim** about the client/server payment join — you're right, `captureOrgEvent`'s `organization_id` + `groups` makes an org-level join possible via #5860; corrected to say that explicitly and left it as a follow-up rather than in scope here.

The report→onboarding attribution part is unchanged from your ✅.

PR #6039, opened 2026-08-13
fix(sandbox): cap the Go daemon's file-edit path at 10MB

pedrofrxncx:
Follows the open PR #5608, which caps the plain-text `read` route at 10MB but leaves the `edit` route (`internal/routes/fs.go`'s `Edit` handler) doing an unbounded `os.ReadFile` on the same untrusted repo files.

A maintainer wants this because the edit route is the same trust boundary as read (any tool call can target an arbitrary file under the repo root), and it's exercised even more often (every write-side tool call), so a large generated file (a log, a lockfile, a big commit) reading unbounded into memory can OOM the daemon and take the sandbox pod down — the daemon's health probe then marks it dead mid-session.

Failure scenario: an agent calls edit on a >10MB text file in the repo; before this fix, `os.ReadFile` loads the whole file (plus the computed replacement string) into memory with no bound. Regression test: `TestEditRejectsOversizedFile` in `fs_edit_test.go` writes an 11MB fixture and asserts a 400 with a "too large" message; `TestEditAllowsFileUnderCap` asserts a normal small edit still succeeds.

To verify: `cd packages/sandbox/daemon-go && go test ./internal/routes/... -run TestEdit -v`.

Locally ran: `gofmt -l`, `go vet ./internal/routes/...`, `go test ./internal/routes/...` — all clean. Full CI validates the rest.

---
## Summary by cubic
Caps the `/_sandbox/edit` route at 10MB to match the existing plain-text `read` cap and prevent OOMs from unbounded file reads. Previously the handler read the whole file regardless of size; now files over 10MB return 400 "File too large" before reading.

• Introduces `maxEditFileBytes = 10 * 1024 * 1024` and a size check in `Edit` (`packages/sandbox/daemon-go/internal/routes/fs.go`).
• Adds `TestEditRejectsOversizedFile` and `TestEditAllowsFileUnderCap` (`packages/sandbox/daemon-go/internal/routes/fs_edit_test.go`).
• Verify with: `cd packages/sandbox/daemon-go && go test ./internal/routes/... -run TestEdit -v`.
• Migration: callers of `/_sandbox/edit` must avoid targeting files >10MB; such requests will receive 400.

<sup>Written for commit 3f3676f2509a3a4e758aaa9756744aa34585e7a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6039?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/daemon-go/internal/routes/fs.go
+++ b/packages/sandbox/daemon-go/internal/routes/fs.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	maxImageBytes    = 5 * 1024 * 1024
+	maxEditFileBytes = 10 * 1024 * 1024
 	maxTransferBytes = 500 * 1024 * 1024
 	transferDeadline = 5 * time.Minute
 )
@@ -417,6 +418,10 @@ func Edit(deps FsDeps) http.HandlerFunc {
 		}
 		if body.OldString == *body.NewString {
 			httpx.Error(w, 400, "old_string and new_string must differ")
+			return
+		}
+		if stat, err := os.Stat(filePath); err == nil && stat.Size() > maxEditFileBytes {
+			httpx.Error(w, 400, fmt.Sprintf("File too large (%d bytes; cap is %d)", stat.Size(), maxEditFileBytes))
 			return
 		}
 		raw, err := os.ReadFile(filePath)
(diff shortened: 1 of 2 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #6060, opened 2026-08-13
feat(sandbox-env): carry the higher memory ceiling on a -medium SandboxTemplate

pedrofrxncx:
## Why

The 4Gi limit was chosen when no pod had ever reached it — the 0.9.14 changelog entry says so explicitly ("0 per-pod OOMs — the 4Gi limit was never reached"). That premise expired on 2026-08-13:

| pod | org | outcome |
|---|---|---|
| `studio-sandbox-prod-xvms5` | `deco-studio` | OOMKilled 17:37Z |
| `studio-sandbox-prod-tdcps` | `deco-studio` | OOMKilled 17:57Z — same thread, same run |
| `studio-sandbox-prod-25nbs` | (other org) | OOMKilled 14:58Z |

`container_memory_rss` for the second one, per minute:

```
17:46  257 MiB   ← boot settles
17:51  242
17:52  314
17:53  667
17:54 2097
17:55 3655       ← 4Gi limit
      → OOMKill
```

RSS, not page cache — a real allocation, inside a `claude-code` agent loop (`cloneOnly` pod, no dev server). The retry path then reprovisioned and reproduced the kill on a second pod, because a limit is not something a retry can route around.

## What

`resources.limits.memory: 4Gi → 6Gi`. Nothing else changes.

**The request is untouched at 2Gi**, which is the whole reason this is a one-line change rather than a capacity decision: the scheduler packs on requests, so this buys no node capacity, makes no warm-pool pod bigger, and costs nothing per pod. It only raises the ceiling one agent loop may burst to.

### The tradeoff, stated in values.yaml

Overcommit. A node packed on 2Gi requests can now hold pods that each reach 6Gi, and node memory-pressure eviction is exactly the failure mode that drove the *request* from 1Gi to 2Gi in 0.9.14. At the observed burst rate — 3 pods in 24h — that risk is small.

If those evictions do come back, the answer is **not** a further blanket raise: it's a separate `SandboxTemplate` so only the agent-loop pods carry the higher ceiling. #5836 already built that machinery (`sandboxTemplateSpec` partial + the pool's real `sandboxTemplateRef` read back off the cluster). It is deliberately not done here, because today it would buy nothing a limit bump doesn't: limits aren't scheduled, so the split's only benefit is bounding overcommit blast radius, and it would cost agent runs their warm-pool binds (a claim naming a template its bound pod wasn't built from never binds) or a second always-on pool of 6Gi idle pods.

### Chart version

0.15.13 → **0.15.14**, deliberately not 0.16.0 — #5836 has claimed that version and is still open.

## Testing

`helm lint` clean. `helm template` with `envName=ci` renders `limits.memory: 6Gi` against an unchanged `requests.memory: 2Gi`; rendering with `selfhost/production/values-sandbox-env-prod.yaml` confirms that sample's own override moved too (it pins the limit, so without the edit it would have silently stayed at 4Gi).

The four places that documented 4Gi were updated with it: the chart README table, the self-hosting production sample, and the en + pt-br Kubernetes docs tables.

## Deploying

Merging does not raise anything by itself — prod pins the chart by `targetRevision` in `deco-apps-cd`, so that needs the bump to 0.15.14 to take effect.

## Related

#6059 makes an OOM kill legible to the user, the agent, and the log — it is what identified this one. Independent of this PR; neither blocks the other.

---
## Summary by cubic
Routes harness-run sandboxes to a new `-medium` SandboxTemplate (3Gi request / 6Gi limit) to avoid 4Gi OOMKills without raising defaults. If `-medium` is missing or unreadable, Studio falls back to the default template so dispatches do not stall.

- Changes
  - Helm (`deploy/helm/sandbox-env`): renders `studio-sandbox-<envName>` and `...-medium` plus matching warm pools; adds `mediumResources`, `warmPool.mediumSize`, and `warmPool.autoscaling.medium` (own HPA with validations); grants RBAC “get” on `sandboxtemplates`; chart `version: 0.15.16`; docs updated (requests now 2Gi; `mediumResources` documented).
  - Studio/provider (`packages/sandbox` and API): introduces `SandboxPurpose` and sends `purpose: "harness-run"` for Claude Code dispatch; derives `cloneOnly` from purpose; resolves the template name via a 5‑minute cached probe and treats 403/404 as “absent” (warns once, degrades to default); persists `purpose` so resurrected claims keep their template; harness-run claims skip tenant pools and select the matching `-medium` warm pool.

- Rollout
  - Recommended: bump `deco-apps-cd` `targetRevision` to `0.15.16` before deploying Studio. Safe if reversed — Studio degrades to the default template with a warning.
  - Optional: set `warmPool.mediumSize` and/or enable `warmPool.autoscaling.medium`.

<sup>Written for commit f9c15fb72c8c59ce708a340c7b7de962d970fe11. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/harnesses/sandbox-dispatch-client.ts
+++ b/apps/api/src/harnesses/sandbox-dispatch-client.ts
@@ -37,7 +37,10 @@ import {
   SANDBOX_GONE_TERMINAL_CODE,
   SANDBOX_UNREACHABLE_PREFIX,
 } from "@decocms/sandbox/dispatch/error-codes";
-import type { SandboxProvider } from "@decocms/sandbox/provider";
+import type {
+  PodTermination,
+  SandboxProvider,
+} from "@decocms/sandbox/provider";
 import { isTransientStreamError } from "@/harnesses/decopilot/built-in-tools/subtask";
 import type { HarnessId, HarnessStreamInput } from "@/harnesses/lib/types";
 import {
@@ -111,12 +114,21 @@ const DAEMON_SILENCE_TIMEOUT_MS = 90_000;
  * the model.
  */
 export class SandboxUnreachableError extends Error {
-  constructor(reason: string) {
+  /**
+   * The bare reason, without the prefix — so a caller that learns WHY the pod
+   * went away can re-wrap without doubling the marker.
+   */
+  readonly reason: string;
+
+  constructor(reason: string, infra?: string) {
     // Prefixed in the CONSTRUCTOR, so the marker `isTransientRunFailure` reads
     // downstream cannot be forgotten by a new throw site. See
     // SANDBOX_UNREACHABLE_PREFIX.
-    super(`${SANDBOX_UNREACHABLE_PREFIX} ${reason}`);
+    super(
+      `${SANDBOX_UNREACHABLE_PREFIX} ${reason}${infra ? ` — ${infra}` : ""}`,
+    );
     this.name = "SandboxUnreachableError";
+    this.reason = reason;
   }
 }
 
@@ -354,10 +366,10 @@ export class SandboxDispatchClient implements SandboxClient {
             virtualMcpId,
             branch,
             sandboxProviderKind: kind,
-            // This pod runs one agent loop and returns its output. Nothing here
-            // opens a preview, so the tenant's install + dev server is pure boot
-            // latency — the harness only needs the checkout.
-            cloneOnly: true,
+            // This pod runs one agent loop and returns its output: no preview,
+            // no dev server, and a memory ceiling of its own (the agent loop is
+            // what OOMKilled 4Gi pods in prod).
+            purpose: "harness-run",
           },
           ctx,
         );
@@ -381,6 +393,13 @@ export class SandboxDispatchClient implements SandboxClient {
         resume: this.resume,
         aborted: () => input.signal?.aborted === true,
         dispatchOnce,
+        lastHandle: () => lastHandle,
+        describeTermination: async (handle) =>
+          handle === null
+            ? null
+            : describeTermination(
+                (await provider.lastTermination?.(handle)) ?? null,
+              ),
       });
     } finally {
       // The run is over — cleanly, failed, or aborted. This pod is `cloneOnly`:
@@ -415,7 +434,9 @@ export class SandboxDispatchClient implements SandboxClient {
  *  - The continuation's own `start` chunk is dropped. `start` RENAMES the message
  *    being folded (AI SDK `processUIMessageStream` assigns `state.message.id`),
  *    so forwarding a second one re-ids a message the projector has already
- *    written parts for. The first `start` of the run wins.
+ *    written parts for. The first `start` of the run wins — but if the interrupted
+ *    attempt died before sending one, the continuation's IS the first and passes,
+ *    or the run's message never opens at all.
  *
  * What it does NOT do is re-run the task: `resume` tells the harness its own
  * context is gone but the work is in the checkout, so it continues from there.
@@ -429,6 +450,18 @@ export async function* dispatchWithContinuation(args: {
   dispatchOnce: (
     resume: { reason: string } | null,
   ) => AsyncIterable<UIMessageChunk>;
+  /**
+   * Ask the infrastructure why the sandbox went away, once it has. Answers the
+   * question the broken stream cannot: an OOM kill leaves no trace in the
+   * daemon's output, so without this both the agent and the user are told only
+   * that the pod "stopped answering" — and a run that OOMs twice reads as a
+   * flaky network instead of a sandbox too small for the work.
+   *
+   * Best-effort: null (or a throw) degrades to the unqualified message.
+   */
+  describeTermination?: (handle: string | null) => Promise<string | null>;
+  /** The handle the failed attempt ran on, for `describeTermination`. */
+  lastHandle?: () => string | null;
   /** Consecutive no-progress dispatches allowed. */
   maxAttempts?: number;
   /** Dispatches allowed in total, progress or not. */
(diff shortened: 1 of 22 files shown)


pedrofrxncx:
Changed approach — the blanket limit raise is reverted in 522d534.

A 6Gi limit over an unchanged 2Gi request means every node packs on 2Gi and may burst to 6Gi, which is the node memory-pressure eviction the request bump to 2Gi (0.9.14) was fixing. So the higher ceiling now rides on a second SandboxTemplate instead:

- The chart renders `studio-sandbox-<env>-medium`, identical to the default except memory: **3Gi request / 6Gi limit** (same 2x overcommit ratio as the default 2Gi/4Gi). `mediumResources` deep-merges over `resources`, so existing overrides still flow. A SandboxClaim has no resource override (CRD: additionalPodMetadata, env, lifecycle, sandboxTemplateRef, warmpool), so a second ceiling has to be a second template.
- A matching `...-medium` SandboxWarmPool at `warmPool.mediumSize` (default 0). An empty pool falls back to a cold pod, so 0 costs nothing idle; it exists so a medium claim has a valid pool name and prewarming later is a one-value change.
- Studio sends `cloneOnly` claims (the Claude Code dispatch path — where all three observed OOMKills were) to that template when `STUDIO_SANDBOX_MEDIUM_TEMPLATE_NAME` is set. Unset = today's behavior, so the chart can land before anything claims it. The claim's `warmpool` follows the template it picked, so a medium claim never binds a 4Gi pod.

Verified by rendering before/after: the default template is byte-identical to main, the medium one differs only in name + memory. `helm lint` clean, 4 new unit tests on the selection, `tsc` clean.

Tradeoff worth a decision at review: dispatch claims currently bind generic warm-pool pods and will now cold-start unless prod sets `warmPool.mediumSize`, at 3Gi reserved per idle replica.

Rollout is still two steps: `deco-apps-cd` `targetRevision` -> 0.15.14, then the Studio env var.

pedrofrxncx:
`STUDIO_SANDBOX_MEDIUM_TEMPLATE_NAME` is gone in b96a043 — the name was always `<STUDIO_SANDBOX_TEMPLATE_NAME>-medium` and the chart always renders it, so the var only bought deploy ordering. Ordering it explicitly is cheaper than a knob that can be left half-set:

**Deploy the chart (0.15.14) before the Studio release that carries this commit.** Studio names `-medium` for `cloneOnly` claims unconditionally now, so in the reverse order every Claude Code dispatch claim fails with `sandboxtemplate not found` until the chart lands.

pedrofrxncx:
Tested against a real cluster (k3s on colima, agent-sandbox operator **v0.4.5**), and reworked the condition. Latest: 8ce46a0.

## Rename: `cloneOnly` → `purpose`

`cloneOnly` was the wrong thing to key on — it's a workload detail, not an identity. `EnsureOptions.purpose` is now `"interactive"` (default) or `"harness-run"`; the dispatch client sets it, `provisionSandbox` *derives* `cloneOnly` from it, and the template + pool selection reads it. It also had to join the persisted ensure-opts, or a resurrected claim would come back on the 4Gi template (test added).

## What the cluster says

Both templates + both pools render and behave:

| case | result |
|---|---|
| interactive claim (default template + pool) | bound the default pool's warm pod — 64Mi/128Mi (test-scaled values) |
| harness-run claim (`-medium` + `-medium` pool) | bound the medium pool's warm pod — 96Mi/192Mi |
| harness-run claim, medium pool at **0 replicas** | cold pod built from the **medium** template — so `warmPool.mediumSize: 0` is safe, which is the default |
| `-medium` template + **default** pool (the mismatch) | did **not** take the default pool's idle pod; cold-created from the medium template. The operator matches warm pods by template hash, so a mismatched pool costs a warm bind, never a wrong ceiling |
| claim naming a **nonexistent** pool | cold pod, no error |
| claim naming a **nonexistent template** | accepted, then parked at `Ready=False reason=TemplateNotFound` **forever** — see below |
| `kubectl scale` on both pools | works (that's the subresource HPA/KEDA drive) |

I corrected two comments that were wrong about the first and fifth rows.

## KEDA, both pools

Installed KEDA on the test cluster and pointed one ScaledObject at each pool: both went `Ready=True Active=True`, KEDA created its HPAs, and both pools' `spec.replicas` moved (default → 2, medium → 1). In-chart, `warmPool.autoscaling.medium` now renders a second HPA against the medium pool (one pool binds one template, so one autoscaler can't straddle both), inheriting unset keys and validated for a metric like the first.

`deco-apps-cd` needs the sibling ScaledObject — prepared on a branch there (`feat/sandbox-medium-pool-keda`, uncommitted): one ScaledObject per env at floor 0, plus a **name-scoped** `ignoreDifferences` entry for the prod medium pool (the existing one is scoped on purpose so tenant pools keep reporting drift). Confirmed the need for it the hard way: `helm upgrade` failed with `conflict with "k3s" with subresource "scale": .spec.replicas` once KEDA owned the field.

## The ordering hazard, now handled in code

Both envs are pinned well behind this chart (**prod 0.12.0, stg 0.10.2**), so "deploy the chart first" is a 3-4 minor upgrade, not a same-hour bump — and the row above shows the failure mode is a claim that hangs until its readiness timeout, per dispatch, fleet-wide.

So the runner now probes the template (one GET, cached 5 min in both directions) and falls back to the default template when it is absent **or** unreadable, warning once. Verified end-to-end against the cluster with the real client code: present → `-medium`, absent → default + warning, interactive → never probes. `403` answers like `404`, which is also why the new `sandboxtemplates: get` RBAC rule can arrive late without breaking anything.

Net effect: merge order no longer matters, and the pin bumps become a normal chart upgrade whenever you want the 6Gi ceiling live.

PR #6087, opened 2026-08-14
fix(file-picker): probe month shards newest-first so fresh uploads surface

guitavano:
## Summary

The image picker (`FilePickerDialog` — "Escolher uma imagem") listed bucket objects "recent first" on page 1 by probing the current/previous **year** prefix (`<prefix>/<yyyy>/`) capped at `MaxKeys`. But S3 `ListObjectsV2` sorts **only lexicographically** — within a year, `01/` sorts before `12/`, so a year-wide probe returns the **oldest** months and truncates the newest ones off the end.

Result: in any active bucket (more than ~50 objects in the current year), a **just-uploaded file never appeared on page 1** — the refetch after upload re-fetched the same truncated window.

Upload keys are `<prefix>/<yyyy>/<mm>/<uuid>-<filename>` (`upload-policy.ts` `buildObjectKey`), so the fix is to shard the probe by **month** and walk backwards from the current month.

## Changes

- New pure helper `monthShardPrefixes(bucketPrefix, now, count)` — generates `<prefix><yyyy>/<mm>/` prefixes newest-first, crossing year boundaries (`2026/01/` → `2025/12/`), mirroring `buildObjectKey`'s UTC `yyyy/mm` shard.
- Page 1 probes those month shards (bounded by `MONTH_SHARD_PROBES = 6`, stops early once the page fills) instead of the two year prefixes. The newest month is fetched directly and then sorted by `lastModified` DESC.
- Broad-prefix fallback (nextCursor + legacy keys) and cursor paging unchanged, now skipping the month shards page 1 already returned.

## Known limitation

If a **single month** holds more than ~50 objects, that month's probe still truncates at `MaxKeys` in UUID-lexicographic (random) order, so a brand-new upload in a very active month could still miss page 1. Eliminating that fully requires a real index with `created_at` rather than relying on S3's lexicographic listing — out of scope here. This PR fixes the cross-month truncation, which is the reported symptom.

## Testing

- Added `describe("monthShardPrefixes")` unit tests: newest-first ordering across the year boundary, zero-padded month matching `buildObjectKey`, UTC derivation, and `count: 0`.
- `bun test apps/api/src/file-storage/file-config-s3.test.ts` — 24 pass
- `bun run --cwd=apps/api check` (tsc) — clean
- `bun run lint` — 0 errors
- `bun run fmt` — applied

---
## Summary by cubic
Shows fresh uploads on page 1 by probing month shards newest-first in parallel. Previously we listed year prefixes; S3’s lexicographic order surfaced the oldest months and hid new uploads in active buckets.

- Adds `monthShardPrefixes` and a shared `monthShardSegment` used by `buildObjectKey` to generate UTC `<yyyy>/<mm>/` shards across year boundaries; page 1 probes up to `MONTH_SHARD_PROBES` months concurrently, then merges, de-dupes, sorts by `lastModified` DESC, and slices to the page size.
- Keeps the broad-prefix fallback to expose `nextCursor`; cursor pages skip month shards and return S3’s lexicographic order (oldest-first).
- Limitation: if a single month holds more than `MaxKeys`, newest items in that month can still be truncated.
- Tests cover `monthShardPrefixes` ordering (UTC, zero-padding, year wrap, multi-year), and shard parity with `buildObjectKey`; typecheck and lint are clean.

<sup>Written for commit 8e0540d85a79067fcaec2065f328af6d41deb726. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6087?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/file-storage/file-config-s3.test.ts
+++ b/apps/api/src/file-storage/file-config-s3.test.ts
@@ -3,6 +3,7 @@ import {
   buildPublicUrl,
   isImageKey,
   matchScanPage,
+  monthShardPrefixes,
   nextScanStep,
   type ScanCandidate,
   stsCredentialProvider,
@@ -105,6 +106,38 @@ describe("isImageKey", () => {
     for (const key of ["a.pdf", "b.docx", "c.png.txt", "folder/", "noext"]) {
       expect(isImageKey(key)).toBe(false);
     }
+  });
+});
+
+describe("monthShardPrefixes", () => {
+  test("lists months newest-first, crossing the year boundary", () => {
+    // 2026-02 walking back 4 months -> Feb, Jan 2026, Dec, Nov 2025.
+    expect(
+      monthShardPrefixes("uploads/", new Date("2026-02-15T00:00:00Z"), 4),
+    ).toEqual([
+      "uploads/2026/02/",
+      "uploads/2026/01/",
+      "uploads/2025/12/",
+      "uploads/2025/11/",
+    ]);
+  });
+
+  test("zero-pads the month to match buildObjectKey's yyyy/mm shard", () => {
+    expect(monthShardPrefixes("", new Date("2026-08-13T18:44:00Z"), 1)).toEqual(
+      ["2026/08/"],
+    );
+  });
+
+  test("uses UTC (keys are UTC-stamped) to derive the month", () => {
+    expect(monthShardPrefixes("", new Date("2026-09-01T00:30:00Z"), 2)).toEqual(
+      ["2026/09/", "2026/08/"],
+    );
+  });
+
+  test("count of zero yields no probes", () => {
+    expect(
+      monthShardPrefixes("p/", new Date("2026-08-13T00:00:00Z"), 0),
+    ).toEqual([]);
   });
 });
 
--- a/apps/api/src/file-storage/file-config-s3.test.ts
+++ b/apps/api/src/file-storage/file-config-s3.test.ts
@@ -9,6 +9,7 @@ import {
   stsCredentialProvider,
 } from "./file-config-s3";
 import type { FileConfigInfo } from "../storage/types";
+import { buildObjectKey } from "./upload-policy";
 
 function info(overrides: Partial<FileConfigInfo>): FileConfigInfo {
   return {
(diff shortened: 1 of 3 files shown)


guitavano:
## Code review pass (multi-perspective) — applied + deferred

Ran a parallel review (correctness / performance / testing-scope). Second commit addresses the actionable findings:

- **Performance (regression I introduced):** month probes now run concurrently (`Promise.all`) instead of sequentially. The sequential early-break only fired when a *single* month held ≥ `target` keys, so sparse/low-volume buckets — the common case — paid all 6 round trips (3 → 7 sequential RTTs) on every picker open. Concurrent probing collapses that to ~1 RTT; a final `sort + slice(target)` keeps the page bounded now that probes over-fetch.
- **Drift guard (testing):** extracted the `<yyyy>/<mm>/` format into `monthShardSegment`, shared by `buildObjectKey` (write) and `monthShardPrefixes` (read), so the write/read shard formats can't silently diverge. Added a parity test + multi-year-wrap + January-wrap cases.
- **Docs (correctness):** the `listObjects` doc claimed cursor pages sort newest-first, but they pass `sort=false` and return S3's lexicographic (oldest-first) order — corrected the comment.

### Deferred (pre-existing, not introduced by this PR)

Both exist identically on `main` today and fixing them means encoding the shard skip-set into the cursor — a larger pagination rework, out of scope here:

1. **Single-object seam drop:** when the month shards fill page 1 (`seen.size == target`), the broad fallback fetches `MaxKeys: 1` only to obtain a continuation token, then breaks before emitting that key — so the single globally-oldest object is skipped between page 1 and page 2. Impact: one object, the oldest, in a newest-first picker.
2. **Month-rollover duplication:** the shard skip-set is derived from `new Date()` per request, not from the cursor, so loading page 1 just before a UTC month rollover and "Load more" just after can re-show the boundary month. Narrow timing window.

Also unchanged and worth naming: `listObjects` orchestration has **no** test in any tier (it hits S3; the e2e stack has no S3/MinIO). The pure `monthShardPrefixes` / `monthShardSegment` helpers are unit-tested.

PR #6207, opened 2026-08-18
fix(mcp): repair invalid tool-call arguments instead of only hinting

pedrofrxncx:
## Summary

Debugging a stuck prod thread (`bagaggio`, Blog Manager agent) surfaced a dead code path in the MCP proxy.

The agent called `POST_EDIT` with `patch` as a JSON *string* and `save: "true"`. The upstream rejected both for -32602. The agent retried the identical shape, gave up on the tool, and re-implemented the call as a sandbox script — ~4M input tokens and several minutes to work around one type mismatch.

Two things were wrong:

**1. The existing hint never fires.** `enrichInvalidParams` only wraps the *throw* path, but the MCP SDK's own server (>= 1.29) catches the validation `McpError` inside its `tools/call` handler and returns an `isError: true` result instead. Every server built on the SDK therefore lands in the shape we don't inspect. Confirmed in prod: no rejection anywhere carries a hint — not in that thread, not in the `spire-agent` logs, which are full of bare `MCP error -32602: Input validation error: …` lines.

**2. Hinting is the wrong ceiling anyway.** -32602 is raised while validating arguments, *before* the tool runs, so a retry is side-effect free. The proxy already holds the tool's advertised `inputSchema` — it can just fix the arguments.

## Changes

- Handle the `isError`-result shape alongside the thrown one (`invalidParamsResultText`, `appendHintToResult`).
- On a -32602 rejection, re-type the arguments against the schema and retry once (`coerceArgsToSchema`). A JSON string for an `object`/`array` parameter, `"true"` for a `boolean`, `"10"` for an `integer`.
- The hint, now only reached when the repair fails, names present-but-wrong-typed arguments. Previously it listed required keys and what was sent — with nothing missing it read as "everything is fine", which is exactly the case that produced this bug.

Coercion is deliberately conservative and fails closed: only string values whose declared type is object/array/boolean/number/integer, only when the string reads cleanly as that type. A `string`-typed parameter is never reinterpreted, an unparseable value is left exactly as sent and the upstream rejection stands, and an `isError` result from a tool that actually ran is never retried.

## Testing

`bun test apps/api/src/mcp-clients/` — 114 pass.

- Unit coverage for `coerceArgsToSchema` (each type, each refusal), `invalidParamsResultText`, `appendHintToResult`, and the new hint clause.
- Three tests in `lazy-client.test.ts` against a **real** `McpServer` over the bridge transport, since the `isError` shape is the SDK server's behavior and a hand-rolled fake would not reproduce it: repair-and-retry succeeds, an unrepairable call gets the hint appended, and an ordinary tool error passes through untouched. The unrepairable case reproduces the exact `POST_EDIT` error text from the thread.

`bun run fmt`, `bun run lint`, and `bun run --cwd apps/api check` are clean.

## Not in scope

The thread's other failures live outside this repo and are noted here so they don't get lost:

- **The actual publish blocker was Perplexity.** `POST_PUBLISH` is gated on a fact-check that returned `skipped_error` on every attempt. That call happens inside `spire-agent`, it is swallowed with no log line, and `factCheckStatus` has not been `ok` anywhere in prod since 2026-07-13 — the failure is invisible until it blocks a publish, and it reads to the user like a content problem. Needs an owner on the Spire side.
- `spire-agent` logs show the mirror-image mismatch on its own outbound calls (`VTEX_IS_GET_PRODUCT_SEARCH`, `query` expected string, got object) — 14 occurrences in 200 minutes. This PR's repair does not cover it: that path sends an object where a string is declared, which is not safely coercible.

---
## Summary by cubic
Repairs MCP tool-call handling and adds a sandbox controller with a new `remote` provider. This stops -32602 retry loops (old: hint that never fired; new: repair-and-retry once, then hint) and moves infrastructure access out of studio without changing preview/daemon traffic.

- MCP argument repair: reads the SDK’s `isError` results as well as thrown errors, coerces string arguments to the declared type from the tool’s `inputSchema` (object/array/boolean/number/integer) and retries once; on failure, appends a clearer hint naming wrong-typed arguments. Safe because validation happens before execution; never retries a tool that ran; leaves unparseable values unchanged.

- Sandbox controller and `remote` provider: introduces a Go controller (`packages/sandbox/controller-go`) and a `remote` `SandboxProvider` (`@decocms/sandbox/server/provider/remote`) that returns the daemon address and token; studio no longer holds kubeconfig or port-forwards. Adds preview resurrection, exports the Kysely state store at `@decocms/sandbox/provider/kysely-state-store`, and mounts a machine-only clone-credential callback under `apps/api` for the controller. Secures both controller and callback by default: require mTLS or `SANDBOX_CONTROLLER_TOKEN`; `SANDBOX_CONTROLLER_INSECURE=1` is the explicit dev opt-out. CI vets/tests/builds the controller in `.github/workflows/sandbox-daemon.yml`.

- Rollout: deploy the controller, set mTLS or `SANDBOX_CONTROLLER_TOKEN`, then enable with `STUDIO_SANDBOX_PROVIDER=remote`. No changes required for existing providers.

<sup>Written for commit a95175b3101643fa5e9c00e0ed6dc70901d3ad7a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:
Branched off the wrong base — reopened clean against main as a 4-file diff.

PR #6238, opened 2026-08-19
feat(jira): two-way sync between Jira boards and the task board

viktormarinho:
## Summary

Opt-in, per-org Jira Cloud integration that mirrors an Agile board into the task
board and pushes agent progress back to Jira. The end state is that Jira becomes
a remote control for the agent: move an issue into a todo column and the Super
Agent picks it up, with status and comments flowing back to the issue.

**Pull** — incremental, driven by a DBOS cron every 10min with a webhook as the
fast path. Issues on the configured board become cards; Jira comments become
card comments. Board membership is `/board/{id}/issue` minus
`/board/{id}/backlog` minus epics (the Backlog tab is not a visible card, and
`standardIssueTypes()` in JQL excludes subtasks but *not* epics), narrowed
further by an optional per-tenant JQL filter. The watermark drives a relative
`updated >= -Xm` JQL with `ORDER BY updated ASC`, capped at 500 issues/run with
a 5min overlap — a truncated run resumes on the next tick rather than skipping
issues.

**Push** — durable, through one queue partitioned per org at concurrency 1, so
posting order is guaranteed. Card status changes resolve a transition by
inverting the tenant's status mapping; card comments post to the issue. Both
directions carry an echo guard (link rows + the integration's own `accountId`),
so a write we pushed is not re-pulled as an inbound change, and re-entrant
workflow steps are idempotent.

**Configuration** is per tenant, against the board's real column names — no
column naming is assumed, so any team's board vocabulary works. Settings →
Tasks holds connect, board select, column mapping, JQL filter, the
`auto_delegate` toggle, enable/sync-now, and the webhook URL with registration
steps.

Field sync (title/description/priority) is pull-only. Credentials are validated
on save via `myself()`; the API token lives in the vault and is never echoed
back. The webhook payload is never parsed — it is a debounced wake-up, and the
secret in the URL is the auth, since Jira admin webhooks are unsigned.

## Notable design points

- **Cards are written as `jira`, not `system`** — `system` is the reports
  importer's identity and freezes title/description.
- **Inbound status only applies when the Jira-side status actually changed**, so
  editing an issue's fields cannot drag a card backwards from where the agent
  moved it.
- **Status push is hooked into `emitTaskBoardUpdated`** — the single funnel for
  every board write — via a dynamic import to break the module cycle, skipping
  writes whose `updatedBy` is `jira`.
- **The initial import never auto-delegates**, otherwise connecting a board with
  N open issues would kick off N agent runs.
- The push queue is registered in `RUN_QUEUES`; without that, a split api/worker
  deploy would enqueue work nothing executes.
- Wiki v2 markup is converted by a real parser (block extractor + per-line rules
  + inline char scanner), not a regex pile, since Jira serves either ADF trees
  or v2 strings depending on the endpoint.

## Migrations

One new migration (`171`), not deployed anywhere yet — it lands with this PR. It
adds `org_jira_integrations` (one per org, vault-backed token, board id/name,
status mapping, JQL filter, `auto_delegate`, DB-generated unique webhook secret,
watermark + last error) and the two link tables that make the echo guard and
idempotency possible.

## Testing

- `bun test apps/api/src/jira` — 17 unit tests (client 8, wiki-markdown 9).
- `bun run check`, `bun run lint` (0 errors), `bun run fmt`, `bunx knip` all
  clean.
- Exercised end to end against a real Jira Cloud board on a local dev org:
  ~200 issues imported as an exact mirror of the board's three columns, 175+
  comments pulled, repeated syncs produce no duplicates, and comment/status
  pushes land on the issue.

Screenshots of the settings page are omitted deliberately — the only instance
configured today points at a customer's board, and this repo is public. Happy to
attach shots from a synthetic board if useful.

## Follow-ups

Known gaps, none blocking the happy path: deletions (a deleted or
backlog-moved issue orphans its card), comment edits/deletes do not propagate,
replies arrive flat, agent screenshots go over as relative URLs, identity
mapping (Jira `accountId` ↔ Studio user), finer 429 handling via
`@decocms/shared/std` `retry`, black-box e2e in `packages/e2e` against a fake
Jira, docs in `apps/docs`, and multi-board/multi-site (the schema already
carries an `id` PK; it needs the per-org UNIQUE relaxed).

Open design questions for review: whether `auto_delegate` should generalize the
existing org-wide auto-assign flag instead of being per-integration, whether
every agent comment should cross over or only milestones, and the conflict
policy when both sides change at once.

---
## Summary by cubic
Adds an opt‑in Jira Cloud integration that mirrors a Jira board to the task board and pushes status and comments back to Jira. Previously there was no Jira sync; now mapped issues appear as cards, card comments post to the issue, and status changes sync both ways with echo guards.

Pull is board‑centric via the Agile API (board issues minus Backlog, epics excluded) with an optional per‑org JQL filter. It runs incrementally by watermark with overlap, orders asc, caps per run, and marks an issue processed only after its comments import; field sync (title/description/priority) is pull‑only. A webhook at POST /api/_jira/webhook/:secret acts as a debounce wake‑up between ~10‑minute cron ticks. Jira wiki v2 is parsed to markdown. Push is durable through a per‑org queue `JIRA_PUSH_QUEUE` (concurrency 1). Status pushes invert the org’s mapping, read the card’s current lane, and verify Jira’s current status before transitioning; comment pushes mirror board comments and are non‑retriable. The echo guard uses link rows and the integration `accountId`. `auto_delegate` assigns the Super Agent via an atomic claim; the initial import never auto‑delegates. `siteUrl` is restricted to Jira Cloud.

**Migrations and rollout**
- Run migration 171 to create `org_jira_integrations` and link tables.
- Ensure workers process `JIRA_PUSH_QUEUE` (added to `RUN_QUEUES`) and register the DBOS Jira sync workflow so pulls run.
- For each org, configure Settings → Tasks → Jira, map statuses, optionally enable `auto_delegate`, and register a Jira webhook to POST to `/api/_jira/webhook/:secret` using the per‑org secret.

<sup>Written for commit 7b819b094f65a5b5a46442e5f33b58d100ca5e7e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


viktormarinho:
Hardening pass on this branch after a review sweep. Everything below is in the
force-pushed single commit; `bun run check`, `lint`, `bunx knip` and
`bun test apps/api/src/jira` (19 tests) are green.

**Fixed — these were the blockers**

- **Plaintext tokens in the DBOS system DB.** The sync cron's work-list step
  returned vault-decrypted integrations, and DBOS persists step output, so every
  enabled org's Jira token was written to that table every 10 minutes. The step
  returns ids now; each leg re-reads and decrypts its own row.
- **SSRF with response echo.** `normalizeSiteUrl` took any dotted host. That URL
  is fetched server-side with the tenant's credentials and the response body
  surfaces in `last_sync_error`, so an `org:manage` principal could read whatever
  the API pod can reach. Restricted to Jira Cloud's own domain, with tests
  covering link-local, private, and look-alike hosts. Data Center support would
  need an explicit allowlist, not a loosening of this.
- **Two paid agent runs on one issue.** Auto-delegate assigned the Super Agent
  with a read-then-write, and the cron, a webhook wake-up (per-pod debounce) and
  a manual sync can all be mid-run on the same issue. Now an atomic conditional
  claim, same pattern as the existing in-review fences.
- **Comment duplicated up to 4x on a customer's issue.** The POST was retriable
  with no idempotency key. Now non-retriable — the comment is already safe in
  Studio, so a missed mirror beats four copies.
- **Issue left in the wrong column, permanently.** The status push trusted the
  lane captured at enqueue, and the enqueue runs after an await, so two quick
  moves could reach the queue out of order and the stale leg would transition the
  issue backwards. The pull can't undo that, because Jira then agrees with
  `jira_status`. It reads the card's current lane now, and asks Jira where the
  issue actually is before transitioning, which also closes the
  crash-between-POST-and-bookkeeping window.
- **Comments stranded forever.** The link's `jira_updated_at` advanced before the
  comment pull, so a failed fetch made the issue look unchanged from then on. It
  advances after the comments land; a new link is claimed at epoch so it still
  wins the UNIQUE against a concurrent run without lying about being processed.
- **First sync never completing.** A successful run over zero issues left
  `last_synced_at` NULL, which pinned the UI on "waiting for the first sync" and
  kept every later run counting as the initial import — the state that suppresses
  auto-delegation, i.e. the headline feature.
- **`enabled` without a board or mapping** was gated client-side only, so the
  cron could error every 10 minutes with nothing to self-heal it. Enforced in the
  tool; changing boards now turns the sync off along with the mapping it clears.
- A failed dynamic import of the Jira module no longer vanishes into an empty
  catch.

**Deliberately deferred** — real, none of them silent or destructive:

- The manual `JIRA_SYNC_RUN` still does the whole import inline in the request,
  so a large first import can outlive the HTTP call and a retry click can start a
  concurrent sync. Should move to the queue.
- The webhook's debounce is per-process, and that path runs the sync outside
  DBOS, so N replicas can sync one integration concurrently. Wasted API calls
  rather than lost data — the UNIQUE constraints hold and the auto-delegate fence
  above closes the expensive consequence. Wants the same wake-up to go through
  DBOS with a dedupe window.
- Column-mapping edits in settings rebuild the whole mapping from cache and
  full-value upsert it, so two fast edits lose the first.
- Rate limiting is still coarse; 429 handling should go through
  `@decocms/shared/std` `retry`.
- `171` collides with the migration number in #6023 — whoever merges second
  should renumber.

PR #6244, opened 2026-08-19
ci: add Depot CI workflows alongside GitHub Actions

viktormarinho:
Adds `.depot/` generated by `depot ci migrate workflows`, which copies `.github/workflows/` into `.depot/workflows/` with Depot-compatibility fixes applied. The `.github/` originals are untouched and keep running on GitHub Actions, so this PR changes nothing about current CI.

## What the migrator did

95 changes across 32 workflows: 60 `runs-on` remaps to `depot-ubuntu-latest`, 25 local-action path rewrites to vendored `.depot/actions/` copies, 7 rewrites of self-referencing path filters, and 3 whitespace-only changes inside folded `if:` expressions.

Verified by parsing both YAML trees and diffing the resolved structures rather than reading the text diff. That mattered: the tool rewrites files via YAML round-trip and in 6 files flattened multi-line `run:` blocks into single escaped strings. The parsed values are identical, so there is no behavior change, but those files are noticeably harder to read. `on:` survived intact and comments were preserved. Both vendored composite actions are byte-identical to their `.github/actions/` originals.

Validation on the committed tree: 30 files, 56 jobs, 312 steps, 0 parse errors, 0 dangling `needs:`, every `./`-relative action resolves.

## Two manual corrections on top of the generated output

**Dropped `native.yml` and `release-native.yaml`.** The migrator remapped their five `macos-latest` jobs to `depot-ubuntu-latest`, flagging it only in a trailing comment. Those jobs build, codesign, and notarize the macOS app, so they cannot pass on Linux. Both workflows stay on GitHub Actions only. Nothing else referenced them except comments.

**Remapped matrix-supplied runners** in `release-studio.yaml` and `preview-build.yaml`. The migrator only rewrites literal `runs-on:` values, so the `ubuntu-latest` and `ubuntu-24.04-arm` entries feeding `runs-on: ${{ matrix.runner }}` in the native multi-arch Docker builds were left as GitHub labels. Remapped to `depot-ubuntu-latest` / `depot-ubuntu-24.04-arm`.

## Not done yet

- **Secrets are not imported.** The 17 secrets and 2 variables these workflows reference are not in Depot CI, so anything needing one fails until they are added. 6 of the 17 are Apple signing material used only by the two dropped workflows.
- **`.depot/` is a second source of truth.** Nothing keeps it in sync with `.github/`. `helm-test.yml` shows the shape of this: its drift check now parses `.depot/workflows/preview-build.yaml`, so on Depot CI it validates the Depot copy while GitHub Actions validates the GitHub copy.
- **`paths-ignore` still lists `.github/**`, not `.depot/**`** in `test.yml`, `e2e.yml`, `resilience.yml`, `multi-pod.yml`, `storage-integration.yml`, `sandbox-daemon.yml`, so editing a Depot workflow triggers the full suite on Depot CI.

Per Depot's docs these workflows only activate once merged to main, so no timing comparison is possible from this PR alone.

---
## Summary by cubic
Adds Depot CI workflows that mirror `.github/workflows/` so CI can run on Depot without changing GitHub Actions, and fixes Playwright installs to use the pinned local binary with a cache keyed to the resolved `@playwright/test` version. Old: CI ran only on GitHub Actions and used `npx playwright install` with a manifest-hash cache; New: Depot-compatible copies under `.depot/` plus `./node_modules/.bin/playwright install` and cache by resolved version. Side effects: `.depot/` is a second source of truth; Depot jobs run only after merge to `main`.

- Mirrored 32 workflows into `.depot/workflows/`: remapped `runs-on` to `depot-ubuntu-*`, rewrote local action paths to `.depot/actions/`, adjusted self-referencing path filters; kept macOS workflows (`native.yml`, `release-native.yaml`) on GitHub Actions; remapped matrix-provided runners in `release-studio.yaml` and `preview-build.yaml`; vendored shared actions (`setup-bun`, `start-minio`, `apt-mirror-fallback`). YAML was reserialized only.
- Stabilized Playwright in `.github/workflows/e2e.yml` and `.github/workflows/test.yml`; regenerated `.depot/` from the fixed sources to match. Cache now keys to the resolved Playwright version; installs call `./node_modules/.bin/playwright`.

- Required to enable green Depot runs: import 17 secrets and 2 variables into Depot; decide how to keep `.depot/` in sync with `.github/`; optionally add `.depot/**` to `paths-ignore` where we already ignore `.github/**`.

<sup>Written for commit 04c5bd79e0ff17377a305ac5dace128033d45f17. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.depot/actions/apt-mirror-fallback/action.yml
+++ b/.depot/actions/apt-mirror-fallback/action.yml
@@ -1,0 +1,100 @@
+name: Run an apt command with Ubuntu mirror fallback
+description: >-
+  Run a command that installs Ubuntu packages, retrying it against a different
+  Ubuntu mirror each time it stalls.
+
+  GitHub's hosted runners point apt at a mirrorlist (`/etc/apt/apt-mirrors.txt`)
+  whose first entry is `azure.archive.ubuntu.com`. When that mirror degrades it
+  does not fail — it crawls (37 kB/s measured on 2026-08-19, so a 21 MB font
+  download could not finish inside 5 minutes), and apt's own mirrorlist fallback
+  only triggers on a hard error, never on a slow transfer. This action gives
+  each mirror a hard wall-clock budget instead: blow the budget, rewrite the
+  mirrorlist to the next mirror, try again.
+
+inputs:
+  run:
+    description: >-
+      Shell command to run. Include `apt-get update` — the mirror changes
+      underneath it between attempts, so its package lists must be refetched.
+    required: true
+  timeout-seconds:
+    description: Wall-clock budget for one mirror's attempt.
+    default: "120"
+  mirrors:
+    description: >-
+      Ubuntu archive mirrors to try, in order, one per line. Every entry must
+      carry the -security pocket as well as the release pocket, because the
+      runner's sources point both at this one mirrorlist.
+    default: |
+      http://azure.archive.ubuntu.com/ubuntu/
+      http://mirrors.edge.kernel.org/ubuntu/
+      http://archive.ubuntu.com/ubuntu/
+
+runs:
+  using: composite
+  steps:
+    - name: Run with Ubuntu mirror fallback
+      shell: bash
+      env:
+        APT_FALLBACK_RUN: ${{ inputs.run }}
+        APT_FALLBACK_TIMEOUT: ${{ inputs.timeout-seconds }}
+        APT_FALLBACK_MIRRORS: ${{ inputs.mirrors }}
+      run: |
+        set -uo pipefail
+
+        # Bound apt's own network waits so a dead connection fails inside the
+        # per-mirror budget instead of consuming all of it.
+        sudo tee /etc/apt/apt.conf.d/99-ci-mirror-fallback >/dev/null <<'APTCONF'
+        Acquire::Retries "2";
+        Acquire::http::Timeout "20";
+        Acquire::https::Timeout "20";
+        APTCONF
+
+        # A timed-out attempt leaves apt-get/dpkg alive holding the dpkg lock,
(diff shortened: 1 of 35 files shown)


viktormarinho:
Closing — we evaluated Depot CI and are not adopting it.

Measured on this branch, same commit on both providers: 43% faster wall clock (8.8 min → 5.0 min) and 33% on the e2e critical path, with the win concentrated in CPU-bound jobs (`build` -36%, `typecheck` -28%) and a ~20s per-job overhead that makes trivial jobs slower (`changes` 5s → 22s).

What ruled it out was compatibility, not speed. Three workflows cannot pass on Depot runners: the sandbox proxy tests bind `::1` and Depot has no IPv6 loopback; `github/codeql-action` needs a GitHub Actions run identity that does not exist off GitHub; and the five macOS jobs have no Depot equivalent at all (the migrator silently remapped them to Linux). Adopting it would mean a hybrid setup plus a permanently duplicated `.depot/` tree with nothing keeping it in sync with `.github/`.

One thing worth keeping came out of it: the cold-cache Playwright install path is broken on main, and Depot's separate cache namespace is what exposed it. Split out into #6259.

PR #6248, opened 2026-08-19
fix(mcp-clients): support JSON Schema union type arrays in the validation hint

pedrofrxncx:
**Source:** bug found auditing `apps/api/src/mcp-clients/validation-hint.ts` after this week's coerce/retry hardening (#6210, #6216, #6242).

**Why a maintainer wants it:** JSON Schema lets a property's `type` be a union array (e.g. a nullable field as `["string", "null"]"`), and the MCP SDK's `Tool.inputSchema.properties` values are untyped custom objects, so nothing stops an upstream server from advertising one. `matchesDeclaredType` and `coerceArgsToSchema` only ever compared against a single type string, so any correctly-typed value against a union-typed property was always reported as "Wrong type" in the repair hint sent back to the agent — actively misleading it about what's actually wrong — and `coerceArgsToSchema` could never recognize a string value as already valid for such a property.

**Failure scenario:** a tool declares `note: { type: ["string", "null"] }`. The agent sends a correct string value but the call still fails validation for an unrelated reason (e.g. a missing required field). The generated hint falsely says `note (expected string,null, got string)` even though `note` was fine, confusing the retry. Regression test: `validation-hint.test.ts` — "does not flag a nullable union type's matching member as wrong-typed" / "flags a value matching none of a union type's members" / "coerces against a union type, skipping when string is already a member".

**Verify:** `bun test apps/api/src/mcp-clients/validation-hint.test.ts` (19 pass).

**Checks run locally:** `bun run fmt`, `cd apps/api && bunx tsc --noEmit` (clean), `bunx oxlint apps/api/src/mcp-clients/validation-hint.ts apps/api/src/mcp-clients/validation-hint.test.ts` (0 warnings/errors), targeted test file above. Full CI validates the rest.

---
## Summary by cubic
Fixes validation and repair hints for tools that declare JSON Schema union types (e.g., ["string", "null"]). Previously, union-typed properties were always flagged as wrong-typed; now we correctly match and coerce across union members.

- Accepts `type` as `string | string[]` and normalizes to an array.
- `matchesDeclaredType` matches any union member (keeps special handling for `"integer"`).
- `coerceArgsToSchema` tries coercion across union members and skips coercion when `"string"` is already allowed.
- `buildValidationHint` formats expected union types as `a|b` and no longer flags values that match a union member.
- Adds targeted tests; no behavior change for single-type schemas.

<sup>Written for commit 1c2adfb308a274ead50ae90e8dbff4d5ec0d6e56. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/mcp-clients/validation-hint.test.ts
+++ b/apps/api/src/mcp-clients/validation-hint.test.ts
@@ -58,6 +58,26 @@ describe("buildValidationHint", () => {
     const hint = buildValidationHint("t", { properties: {} }, { a: 1 });
     expect(hint).toBe('Invalid arguments for "t". You sent: a.');
   });
+
+  it("does not flag a nullable union type's matching member as wrong-typed", () => {
+    const hint = buildValidationHint(
+      "t",
+      { properties: { note: { type: ["string", "null"] } } },
+      { note: "hi" },
+    );
+    expect(hint).not.toContain("Wrong type");
+  });
+
+  it("flags a value matching none of a union type's members", () => {
+    const hint = buildValidationHint(
+      "t",
+      { properties: { note: { type: ["string", "null"] } } },
+      { note: 5 },
+    );
+    expect(hint).toContain(
+      "Wrong type: note (expected string|null, got number).",
+    );
+  });
 });
 
 describe("enrichInvalidParams", () => {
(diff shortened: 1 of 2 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #6215, opened 2026-08-19
fix(sandbox): cap the Go daemon's file-read path at 10MB

pedrofrxncx:
Follows the same bug class as open PR #6039 (`/edit` file-read cap), but on a different route: `Read`'s text path (`packages/sandbox/daemon-go/internal/routes/fs.go`).

`/read` already caps its image path at `maxImageBytes` (5MB), but for a text file it falls through to an unconditional `os.ReadFile(filePath)` regardless of size. A large text file (a log, a generated artifact, an oversized decofile) gets read fully into memory with no bound, which can OOM the daemon and tear down the sandbox pod on the next missed health probe — the exact failure mode the file's own `decodeBody` comment already documents for the request-body side.

Added a `maxTextReadBytes` (10MB, matching the cap #6039 is adding to `/edit`) check right before the read, returning a 400 with the file size and cap instead of buffering the whole file.

Failure scenario: an agent runs `/read` against a >10MB text file in the sandbox → daemon buffers the entire file into memory → repeated/concurrent large reads can push the daemon over its memory limit and get the pod killed mid-session.

Regression test: `TestReadRejectsOversizedTextFile` (fixture file just over the cap, expects 400 with a "too large" message) and `TestReadAllowsTextFileUnderCap` (small file still reads fine) in `fs_test.go`.

Reviewer check: `cd packages/sandbox/daemon-go && go test ./internal/routes/...`

Locally ran: `go build ./...`, `go test ./internal/routes/...`, `gofmt -l` (clean), `go vet ./internal/routes/...` (clean). Full CI validates the rest.

---
## Summary by cubic
Caps text-file reads in the sandbox Go daemon’s `/read` route at 10MB to prevent OOMs. Previously, the text path used unbounded `os.ReadFile`; now requests for files >10MB return 400 with the file size and cap. This aligns with the `/edit` 10MB cap.

**Review notes**
- Adds `maxTextReadBytes` and a pre-read size check in `packages/sandbox/daemon-go/internal/routes/fs.go`.
- Adds `TestReadRejectsOversizedTextFile` and `TestReadAllowsTextFileUnderCap` in `packages/sandbox/daemon-go/internal/routes/fs_test.go`.
- Verify with: `cd packages/sandbox/daemon-go && go test ./internal/routes/...`

<sup>Written for commit 134f0f2c2485c2139370ad2a570c62ff8d88d3f4. Summary will update on new commits.</sup>

[...]

PR #6252, opened 2026-08-19
fix(sandbox): cap the Go daemon's config-update body size

pedrofrxncx:
**Source:** bug found while auditing the Go sandbox daemon (`packages/sandbox/daemon-go/internal/routes`) for reliability gaps — the same class already fixed for the file-transfer routes (`fs.go`'s `decodeBody`, see #6215/#6039).

**Payoff:** `ConfigUpdate` (`PUT /_sandbox/config`) and `OrgFsConfig` both read the request body with a raw, unbounded `io.ReadAll(r.Body)`. A misbehaving or malicious caller can stream an unbounded body into memory and crash the daemon — this is a single static binary per sandbox pod, and Studio tears the pod down on the next missed health probe, so an OOM here kills the whole session.

**Failure scenario:** POST a multi-GB body to `/_sandbox/config` or the org-fs config route → `io.ReadAll` buffers it all into memory with no limit → daemon OOMs/crashes → pod dies. `fs.go` already guards its routes against exactly this with a capped `decodeBody`; these two routes were missed.

**Fix:** added a shared `readLimitedBody(r, maxBytes)` helper in `util.go` (same pattern as `fs.go`'s `decodeBody`, just returning raw bytes since these two callers each unmarshal differently) and wired both routes through it, capped at 1MB — config/org-fs payloads are always small hand-written JSON, never file transfers, so 1MB is generous headroom not a real ceiling.

**Regression test:** `TestReadLimitedBodyRejectsOversizedRequest` in `util_test.go`, mirroring the existing `TestDecodeBodyRejectsOversizedRequest` in `fs_test.go` — drives an infinite reader through the helper and asserts it's rejected instead of buffered.

**To confirm:** `cd packages/sandbox/daemon-go && go build ./... && go vet ./... && go test ./internal/routes/...`

**Locally verified:** `go build`, `go vet`, `gofmt -l` (clean), and the full `internal/routes` test package (including the existing config-redaction tests) — all green. CI validates the rest.

---
## Summary by cubic
Caps the request body for config endpoints at 1 MB to prevent daemon OOMs. Previously, `ConfigUpdate` and `OrgFsConfig` read unbounded bodies into memory; now oversized bodies are rejected with 400.

- Adds `readLimitedBody` in `internal/routes/util.go` and applies it to `ConfigUpdate` and `OrgFsConfig`.
- Sets `maxConfigBytes` to 1 MB; normal small JSON payloads are unaffected.
- Aligns config routes with the existing `fs.go` protection pattern.
- Adds `TestReadLimitedBodyRejectsOversizedRequest` to prevent regressions.

<sup>Written for commit de5fff6bb04ae7f1aa605d0a2bcd09e6e3912be3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6252?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sandbox/daemon-go/internal/routes/config.go
+++ b/packages/sandbox/daemon-go/internal/routes/config.go
@@ -2,7 +2,6 @@ package routes
 
 import (
 	"encoding/json"
-	"io"
 	"net/http"
 	"sort"
 	"strings"
@@ -99,7 +98,7 @@ func ConfigRead(deps ConfigDeps) http.HandlerFunc {
 
 func ConfigUpdate(deps ConfigDeps) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		raw, err := io.ReadAll(r.Body)
+		raw, err := readLimitedBody(r, maxConfigBytes)
 		if err != nil {
 			httpx.Error(w, 400, "bad body: "+err.Error())
 			return
--- a/packages/sandbox/daemon-go/internal/routes/misc.go
+++ b/packages/sandbox/daemon-go/internal/routes/misc.go
@@ -1,7 +1,6 @@
 package routes
 
 import (
-	"io"
 	"log/slog"
 	"net/http"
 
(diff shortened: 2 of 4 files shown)


pedrofrxncx:

[studio-bot:stale-close]

PR #6256, opened 2026-08-19
ci: survive a degraded Ubuntu mirror in the Playwright/apt steps

tlgimenes:
## What broke

`azure.archive.ubuntu.com` — the first entry in the hosted runner's `/etc/apt/apt-mirrors.txt` — degraded to **~37 kB/s**. `playwright install-deps chromium` downloads ~21 MB of font packages, which cannot finish inside the step's 5-minute budget at that rate.

apt's own mirrorlist fallback never fired: it triggers on a *hard error*, never on a *slow transfer*. So every attempt sat on the same crawling mirror until the step timed out.

Failing runs on `main`: [32278546205](https://github.com/decocms/studio/actions/runs/32278546205) (E2E, shards 1 and 2), [32278546215](https://github.com/decocms/studio/actions/runs/32278546215) (web-component-tests).

```
Get:2 http://azure.archive.ubuntu.com/ubuntu noble/universe amd64 fonts-ipafont-gothic all [3513 kB]
##[error]The action 'Install Playwright chromium system deps' has timed out after 5 minutes.
```

(Shard 2 did drift to `archive.ubuntu.com` mid-update and still timed out — by then most of the budget was already gone.)

## The fix

**New composite action `.github/actions/apt-mirror-fallback`.** It gives each mirror a hard wall-clock budget; when the budget blows it rewrites the mirrorlist to the next mirror and retries. Mirrors, in order:

| mirror | why |
| --- | --- |
| `azure.archive.ubuntu.com` | in-region, normally the fastest |
| `mirrors.edge.kernel.org` | Fastly-backed, independent of Azure |
| `archive.ubuntu.com` | Canonical's own |

All three serve the `-security` pocket, which matters because the runner points both its release and security sources at this single mirrorlist. One mirror per attempt, deliberately — leaving the others in the file would let apt silently fall back *inside* an attempt, which is exactly the behaviour that fails to trigger on a merely-slow mirror. It also bounds apt's own connect/read timeouts, and kills + `dpkg --configure -a`s the apt state a timed-out attempt leaves behind so the next mirror doesn't fail on a stale lock.

**Split the browser download off the deps install.** The chromium binary comes from Playwright's CDN and genuinely must succeed, so it no longer rides behind apt via `--with-deps`.

**The deps step is `continue-on-error`.** Every shared library chromium links against is already in the runner image (the failing log shows all of them as `already the newest version`) — the only thing that step actually downloads is fonts, and there are no screenshot assertions in either suite. A total mirror outage should cost glyph coverage, not a red build. A genuinely missing library still surfaces loudly, as a chromium launch failure in the tests themselves.

**Same action for the sandbox-daemon `ripgrep` install**, which had the identical exposure.

**`workflow_dispatch` on the three workflows.** Their push *and* PR triggers both skip `.github/**`-only changes, so a CI-infrastructure fix had no way to validate itself. (Takes effect once this is on `main`.)

## Testing

- Extracted the action's shell body and ran it under GitHub's exact shell (`bash --noprofile --norc -eo pipefail`) with stubbed `sudo`/`timeout` across four cases: all mirrors fail → exit 1 with `::error::`; third mirror succeeds → exit 0 after two `::warning::`s; first mirror succeeds → exit 0, no retries; empty mirror list → exit 1. All as expected.
- All four YAML files parse.
- Verified all three mirrors serve `dists/noble-security/Release` (`200`, 0.5 s / 1.1 s / 2.5 s).
- The PR carries a **`TEMP:` commit** adding `apps/web/.ci-validation-marker` purely to flip the `changes` gate so this PR's own CI exercises the changed steps. **It must be dropped before merge** — I'll remove it once CI is green.

## Affected areas

CI only. No application code changes.

---
## Summary by cubic
Hardens CI against slow Ubuntu mirrors in apt-backed steps. Previously `playwright install-deps chromium` and `apt-get install` could stall on `azure.archive.ubuntu.com` and time out; now each attempt has a wall-clock budget with mirror fallback, and the Chromium download no longer rides behind apt.

- New composite action `.github/actions/apt-mirror-fallback`: time-boxes each attempt, rotates `/etc/apt/apt-mirrors.txt` across `azure.archive.ubuntu.com` → `mirrors.edge.kernel.org` → `archive.ubuntu.com`, bounds apt connect/read via `Acquire::*` timeouts, and cleans dpkg locks between retries.
- Split Playwright steps: `npx playwright install chromium` runs alone; `npx playwright install-deps chromium` runs via the fallback action and is `continue-on-error` (fonts may be skipped; missing libs still fail at runtime).
- Route `ripgrep` install in the sandbox-daemon workflow through the same action.
- Add `workflow_dispatch` to `e2e`, `test`, and `sandbox-daemon`; bump the web component test job timeout to 20 minutes for worst-case fallback.
- CI-only change; no app code affected.

<sup>Written for commit 767e3a17fb981a0f9d90b828aa80b3186284d462. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6256?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/actions/apt-mirror-fallback/action.yml
+++ b/.github/actions/apt-mirror-fallback/action.yml
@@ -1,0 +1,100 @@
+name: Run an apt command with Ubuntu mirror fallback
+description: >-
+  Run a command that installs Ubuntu packages, retrying it against a different
+  Ubuntu mirror each time it stalls.
+
+  GitHub's hosted runners point apt at a mirrorlist (`/etc/apt/apt-mirrors.txt`)
+  whose first entry is `azure.archive.ubuntu.com`. When that mirror degrades it
+  does not fail — it crawls (37 kB/s measured on 2026-08-19, so a 21 MB font
+  download could not finish inside 5 minutes), and apt's own mirrorlist fallback
+  only triggers on a hard error, never on a slow transfer. This action gives
+  each mirror a hard wall-clock budget instead: blow the budget, rewrite the
+  mirrorlist to the next mirror, try again.
+
+inputs:
+  run:
+    description: >-
+      Shell command to run. Include `apt-get update` — the mirror changes
+      underneath it between attempts, so its package lists must be refetched.
+    required: true
+  timeout-seconds:
+    description: Wall-clock budget for one mirror's attempt.
+    default: "120"
+  mirrors:
+    description: >-
+      Ubuntu archive mirrors to try, in order, one per line. Every entry must
+      carry the -security pocket as well as the release pocket, because the
+      runner's sources point both at this one mirrorlist.
+    default: |
+      http://azure.archive.ubuntu.com/ubuntu/
+      http://mirrors.edge.kernel.org/ubuntu/
+      http://archive.ubuntu.com/ubuntu/
+
+runs:
+  using: composite
+  steps:
+    - name: Run with Ubuntu mirror fallback
+      shell: bash
+      env:
+        APT_FALLBACK_RUN: ${{ inputs.run }}
+        APT_FALLBACK_TIMEOUT: ${{ inputs.timeout-seconds }}
+        APT_FALLBACK_MIRRORS: ${{ inputs.mirrors }}
+      run: |
+        set -uo pipefail
+
+        # Bound apt's own network waits so a dead connection fails inside the
+        # per-mirror budget instead of consuming all of it.
+        sudo tee /etc/apt/apt.conf.d/99-ci-mirror-fallback >/dev/null <<'APTCONF'
+        Acquire::Retries "2";
+        Acquire::http::Timeout "20";
+        Acquire::https::Timeout "20";
+        APTCONF
+
+        # A timed-out attempt leaves apt-get/dpkg alive holding the dpkg lock,
+        # which would fail the next mirror for the wrong reason. Kill and repair.
+        reset_apt() {
+          sudo pkill -9 -x apt-get 2>/dev/null || true
+          sudo pkill -9 -x dpkg 2>/dev/null || true
+          sudo dpkg --configure -a >/dev/null 2>&1 || true
+        }
+
+        # The mirrorlist is a GitHub-hosted-image thing: on any other runner the
+        # sources do not point at it, so switching mirrors is a no-op and only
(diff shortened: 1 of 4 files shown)


tlgimenes:
## Validated, then the `TEMP:` commit was dropped

Full CI ran green against commit `f48d87660` (fix + the temporary `apps/web/.ci-validation-marker` that flipped the `changes` gate):

| job | result |
| --- | --- |
| `web-component-tests` | pass — this is the job that was timing out on `main` |
| `e2e-shard (1)` / `e2e-shard (2)` | pass |
| `test`, `typecheck`, `lint`, `format`, `build` | pass |

From [e2e-shard (1)](https://github.com/decocms/studio/actions/runs/32280689778), the deps step behaved exactly as intended — mirrorlist found, first mirror served the whole payload, no fallback needed, zero warnings:

```
##[group]apt via http://azure.archive.ubuntu.com/ubuntu/
0 upgraded, 9 newly installed, 0 to remove and 20 not upgraded.
Need to get 21.1 MB of archives.
apt succeeded via http://azure.archive.ubuntu.com/ubuntu/
```

13 s for the same 21.1 MB that could not finish in 5 minutes earlier today — azure has recovered, which is why the fallback path did not need to fire. The fallback path itself is covered by the offline test run described in the PR body (all mirrors fail / third succeeds / first succeeds / empty list), run under GitHub's exact `bash --noprofile --norc -eo pipefail`.

The `TEMP:` commit has now been force-pushed away; the branch is the single `ci:` commit. The remaining checks on this head will legitimately skip, since a `.github/**`-only change is what the `changes` gate is designed to skip.

Not covered by this PR's own CI: the `ripgrep` call site in `sandbox-daemon.yml` (its gate needs a `packages/sandbox` change). Same action, already exercised by the two Playwright call sites.

PR #6401, opened 2026-08-21
fix(threads): sanitize object keys, not just values, before storing a part payload

pedrofrxncx:
Follows the `thread_message_parts` NUL/lone-surrogate hardening (already-merged, see `serializePayload` in `apps/api/src/storage/thread-message-parts.ts`) with a gap I found while auditing that path per this tick's focus area.

**Bug:** `serializePayload`'s `JSON.stringify` replacer strips NUL bytes and lone UTF-16 surrogates from string *values* so a part payload never trips Postgres's SQLSTATE 22P05 ("unsupported Unicode escape sequence") on INSERT. But a JSON replacer function can only substitute what a key maps to — it never gets to rename the key itself. So an object key that carries a NUL byte or an unpaired surrogate (e.g. a tool result keyed by attacker/content-derived strings, not just its values) still serializes to a raw `\u0000` escape sequence and hits the exact same 22P05 this code exists to prevent. Per the file's own comment, that failure strands the whole run `in_progress` forever, since the projection step is the sole writer of terminal thread status.

**Fix:** adds `sanitizeKeysForPg`, which rebuilds the payload tree with clean keys before stringifying (values still pass through the existing replacer). Normal keys are byte-identical output — both regexes are no-ops on clean input — so ids derived from the payload stay stable, matching the existing guarantee for values.

**Test:** two new cases in `apps/api/src/storage/thread-message-parts.test.ts` (co-located unit test, no DB needed) — a NUL byte in a key and a lone surrogate in a key — mirroring the existing value-sanitization tests in the same file.

Reviewer check: `cd apps/api && bun test src/storage/thread-message-parts.test.ts`

Locally verified: `bun run fmt`, `bunx tsc --noEmit` (apps/api), the targeted test file above (10/10 pass), and `bunx oxlint` on both touched files (0 errors, 1 pre-existing unrelated warning on the NUL regex literal). Full CI validates the rest.

---
## Summary by cubic
Sanitizes JSON object keys in thread part payloads before storage to prevent Postgres SQLSTATE 22P05. Previously only string values were sanitized; keys with NUL bytes or lone surrogates caused INSERT failures that left runs in_progress.

- Introduces sanitizeKeysForPg to rebuild the payload with cleaned keys; the existing JSON.stringify replacer still sanitizes string values.
- Clean inputs are byte-identical; payload-derived IDs remain stable.
- Adds unit tests for NUL and lone surrogate in keys. Reviewer: cd apps/api && bun test src/storage/thread-message-parts.test.ts
- No migrations or API changes.

<sup>Written for commit 14fdad300178a4e8b3bb50182ee41fbf49aa0074. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:

[studio-bot:stale-close]

PR #6541, opened 2026-08-25
fix(organization): drop the redundant id input from org update/delete tools

Tyagiquamar:
## Problem

`ORGANIZATION_UPDATE` took `id` as the org to update but never checked it against `ctx.organization?.id` (the path-resolved, membership-verified org). A caller authenticated against org A could pass `input.id` for org B and reach `boundAuth.organization.update` with an org they may have zero membership in. As with ORGANIZATION_DELETE before #6501, Better Auth's own `updateOrganization` re-checks permission for the target org, so this is not a live bypass today — but it left the tool one vendored-dependency behavior away from one.

## Fix

Per review, the parameter is gone rather than guarded: both `ORGANIZATION_UPDATE` and `ORGANIZATION_DELETE` drop `id` from their input schemas and take the org from `requireOrganization(ctx)`, which is already resolved from the URL path — so there is nothing left to compare and nothing to spoof. `tools-rest.ts` parses with a non-strict `z.object`, so any agent still sending `id` keeps working and just gets it stripped. The shared `tool-io.ts` types and the settings UI's `ORGANIZATION_DELETE` call were updated to match.

## Testing

- `bun test apps/api/src/tools/organization/update.test.ts apps/api/src/tools/organization/delete.test.ts` → **6 pass, 0 fail** (cases now assert the update targets the org resolved from the request context).
- `tsc --noEmit` on `packages/shared`, `apps/api`, `apps/web` → exit 0.
- `biome format` on touched files → clean.

---
## Summary by cubic
Drops the `id` input from `ORGANIZATION_UPDATE` and `ORGANIZATION_DELETE`, so the target organization always comes from the URL-resolved request context instead of client-supplied input. `ORGANIZATION_DELETE` is now idempotent: deleting an already archived organization returns success without updating it.

- The organization is now taken via `requireOrganization(ctx)`, leaving no id to compare or spoof.
- Clients that still send `id` keep working because `tools-rest` parses non-strictly and strips the field.
- The settings UI's delete call and shared `tool-io` types were updated to match.

<sup>Written for commit 91969abb205824ba3d1a435780ab0ad448cad9a9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/6541?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #7116, opened 2026-09-09
feat(plans): tiered plans, feature gates and the AI usage bar

pedrofrxncx:
The studio half of tiered plans. Gateway half: **decocms/ai-gateway#30, which should merge first** — without it `DECO_AI_GATEWAY_ENABLED` is unset, no gateway adapter exists, and every gate here fails open, so this PR is inert rather than broken.

> This branch now carries the design pass that was briefly opened as #7218 (closed in favour of this one, which is where the work originated). The server-side half below is unchanged; the ten `apps/web` commits on top — plus four lines of `packages/ui/src/styles/global.css` — are @rafavalls's, and are described under [The design pass](#the-design-pass).

## The split this PR preserves

**Studio stays open source and holds no pricing logic.** It hardcodes the feature *keys* — `cms`, `chat`, `kanban`, `monitoring`, `model_choice`, `diagnostic`, `diagnostic_enriched`, all public strings — and every *value* is the gateway's answer for that org at that moment. That is what lets a gate or an allowance move without a mesh deploy. No allowance, price or margin appears anywhere in this diff, with one marked exception (`planPriceBrl`, below).

## Server-side enforcement

`requiresFeature` on `defineTool` sits in the same execute wrapper as the org-block gate, so **declaring it is the gate** — no handler remembers a check. The surfaces that are *not* tools each declare their own at their own single chokepoint, which is one insertion per surface because each has exactly one:

| Feature | Where it runs | Gate |
|---|---|---|
| `diagnostic_enriched` | `COMMERCE_DISCOVERY_RUN` | `requiresFeature` (+ `requiresAiBudget`) |
| `chat` | the Decopilot `POST …/threads/:id/messages` route | `orgHasFeature` + `assertAiBudget` at the top of the handler |
| `monitoring` | the CDN warehouse via a BFF `fetch` | `requireOwnedSite` — the one guard all four monitor routes share |
| `cms` | the sandbox-less Fast Preview data plane | `resolveDecofileScope` — the one middleware all four decofile routes share |

A plan refusal returns **403 with a `code`** and is logged as a warning, not through the error path — it is the expected answer for an org without the feature, not an incident.

**Every gate fails OPEN** when the gateway has no answer at all (no gateway configured, no user to mint a JWT for, or a gateway that could not be reached with nothing cached). This is product gating, not access control: the gateway is what actually refuses the spend, and locking a paying org out of its CMS because one fetch blipped is the worse bug. A *refresh* failure serves the stale entry rather than denying.

**Known gap:** the sandbox proxy (`/api/:org/sandbox/*`) stays open, deliberately — it is the whole sandbox surface (files, exec, preview, events) and not CMS-specific, so gating it on `cms` would lock agents and dev sessions out. An org without `cms` can still reach content through a live sandbox.

## The exhausted-bar stop — built, dormant

`assertAiBudget` refuses once the bar reads `exhausted`; `requiresAiBudget: true` is the declarative form. This is the margin stop the bar was otherwise only *reporting*.

- **Behind `STUDIO_PLAN_USAGE_ENFORCED`, default off.** A hard refusal on the chat and dispatch hot paths must not arrive with a deploy, and a self-hosted deployment never wants it.
- **`usage: null` is unknown, never exhausted.** A gateway that could not read consumption must not stop a paying org's work. Unit-tested.
- **Opt-in per surface, not a blanket stop.** `cms` and `monitoring` declare no budget, so a full bar leaves them working — which is the promise the billing card makes ("CMS keeps working, chat pauses").

The Kanban gate and the budget stop apply at the task-dispatch paths and the two sandbox routes that spend, not just the board tools.

## Model governance

Model choice is Ultra-only, and **the gate covers the model's name, not just the picker** — below Ultra an org is never told which model ran its chat, Kanban task, image generation or web research.

- **`resolveTier` now resolves pin → user override → org slot → default-pick.** A deco-admin pin from the gateway outranks everything, but *only* for an org without `model_choice`; an Ultra org bought the right to choose and keeps it. So below Ultra an org's writes to its own slots stopped deciding anything.
- **`AI_PLAN_ENTITLEMENTS` projects the pin out** through `toClientEntitlements` — an explicit **allowlist**, not a `delete`, so a field added to the wire later is omitted by default instead of leaking on the next gateway deploy. Unit-tested for exactly that, including that the payload contains no model name.
- **The org model-slot table is withheld below Ultra**, not disabled: a disabled row still reads the names out. Safe to hide only because the pin (or the existing default-pick) supplies the model — an org that never saved a slot has always worked.
- **Six sites that rendered a dollar figure or a model name for every org** now read `useShowThreadCost()`. The cost and model rows *drop out* below Ultra rather than greying — a greyed "cost —" still says a cost exists. Token counts are unaffected, including the in/out rows that used to be nested inside the `cost > 0` branch. One of the six was an `aria-label`, where a screen reader leaked what the eye did not.
- **Entitlements are not persisted to localStorage**, so a stale cached decision cannot paint paywalls or model names.

## Client gates and the paywall

`useFeature("<key>")` answers one gate; `<FeaturePaywall feature="…">` is the popup it opens on. Also fails open — in flight and on a failed read, every feature reads as included.

| Feature | Gate | Shape |
|---|---|---|
| `cms` · `kanban` · `monitoring` | `TabBody`, keyed by `featureForTab` | Popup over a closed panel |
| `chat` | `handleSubmit` in `chat/input.tsx` | Popup **instead of** the send — gated at the submit rather than on `canSubmit` so the click explains itself; a dead button reads as "my message failed", not "my plan is why". The draft is kept |
| `model_choice` | `TierTrigger` | **No popup** — the name is withheld: no cog, and the row falls back to its blurb |

Gating at the tab *body* rather than the tab *button* is deliberate: a deep-linked `?main=board` goes through the same door.

## The design pass

- **One tier ladder.** `plan-ladder.tsx` is the single answer to "what does this tier look like" — its mark, its colour, its price. The plan card, the catalog and the paywall all read from it, so they cannot disagree.
- **One paywall dialog.** `PaywallDialog` is the shell; `FeaturePaywall` fills it in. The gated panel renders a populated, inert board behind it so the dialog is not sitting on an empty screen. Site Editor and Monitor deliberately have no backdrop — see Known gaps.
- **Locks in the sidebar.** A row whose view the plan withholds carries a lock, defined once in `use-tab-locked.ts` so both nav lists agree.
- **One top-up picker.** The zero-balance and exhausted-mid-stream dialogs were byte-for-byte copies; `TopUpAmounts` is the shared one, under a new `credits` i18n domain.
- **A Free org is told why.** Both credits dialogs bailed with `return null` on a plan without `credits`. The empty state had no fallback, so an org whose allowance was spent opened a chat it could not run and was told nothing. It now shows the plan dialog.

## The billing card

Plan name, a **Change plan** button, the bar as a percent with ok/warn/exhausted colouring, and "Resets on 8 October 2026" — a real date, from the same rule that performs the roll, not "next month". Plans with no AI envelope hide the bar entirely and say so; a full red bar would read as *you burned through it* on an org that never had any. At 100% a second line appears with the credit balance in dollars — the only amount in the card, and only at the moment credits are what the org is spending. The usage bar is a percent, never a dollar, and advances optimistically by each turn's own cost.

The card renders an explicit "couldn't read this organization's plan" with a retry instead of returning `null`, and `useEntitlements` no longer swallows the error. A failure the user cannot see is a failure nobody can report — previously a 500 and a self-hosted deployment with no plan surface looked identical: an absent card.

## Payment

Stripe is the only way to acquire a paid tier: a price mapped in `STRIPE_PLAN_PRICE_IDS` grants the tier on an active subscription and revokes it on cancellation. Studio refuses to boot with plans on and no explicit gateway JWT secret or provision key, rather than silently opening every gate.

## Needs a decision before merge

- **Prices are hardcoded** in `planPriceBrl`, keyed by plan id, until Stripe checkout is wired end to end. This is the one place a number appears in the diff, and it is marked temporary. Keyed by id and not by rung on purpose: a mark on the wrong rung is cosmetic, a price on the wrong plan is a misquote.
- **`Pro+` wants to be `Scale`.** Product decision made; needs a gateway migration (`drizzle/0009_add_plans.sql` seeds the name), so it is not in this diff.
- **Below Ultra, gated UI disappears silently** — the tier pill in the composer, Default models and the BYO tiles on Billing, the Cost card on Monitor. Consistent with itself, but it contradicts the sidebar, which locks rather than hides. Worth settling in one pass across all three.
- **The Kanban trial has no client, and it collides with a quota mesh already has.** `POST /api/teams/:org/tasks/claim` exists on the gateway with e2e coverage and mesh never calls it. `apps/api/src/billing/task-quota.ts` is already a task-execution quota with the *same* shape — claim keyed by task id, charged at dispatch, refunded when a run produced nothing, period keys without cron, per-org overrides, behind `STUDIO_TASK_QUOTA_ENFORCED` — but it answers "has this org paid for task executions?" against the **Stripe subscription**, where the gateway's grant answers "does this plan include kanban, or has it trial grants left?" against the **plan**. Wiring both means two sources of truth and two refund paths. The plan should eventually subsume the subscription quota — that is a migration, not a wiring task.

## Known gaps

- **No paywall backdrop for CMS or Monitor.** Kanban's works because `TaskBoardPage` reads one plain query key. Site Editor is a live sandbox iframe with nothing to seed; the CDN tab depends on a suspense query keyed by a runtime MCP client object plus ~7 derived keys and default component state. Left blank rather than coupled to two components' internals.
- **A downgrade orphans BYO provider keys.** `ConnectedProvidersSection` is gated on `model_choice`, so an org that drops from Ultra keeps its keys in the database with no UI to see or remove them. Not fixed here.
- **The subscribe error toast leaks a dev string** (`Stripe API 503: plan 'pro' has no Stripe price configured`).
- **`ORGANIZATION_SETTINGS_UPDATE` is still writable below Ultra.** It is now a write that *does nothing* (the pin outranks it) rather than a privilege escalation, which is why this is a note and not a blocker — but a `requiresFeature` there would make it say so.
- **No authorization for spending.** Plan changes are guarded by the same `ctx.access.check()` the sibling AI-provider tools use, which is not a "may this person commit the org to money" check. Any member can move the org to Ultra. Must tighten alongside payment.
- **`etag` buys nothing yet.** The gateway emits it and the adapter types it, but there is no `If-None-Match`; the card leans on react-query's 60s `staleTime`. Both response shapes now go through one `toPlanEntitlements` mapper, which is what a revalidation change would hook into — they had drifted as two copies.

## Testing

- `bun run check`, `bun run lint` and `bun run fmt` pass.
- `plan-feature-gate.test.ts` covers both pure decisions: `isFeatureAllowed` (absent = denied, `null` = fail open, a truthy non-boolean is not granted) and `isUsageBlocked` (dormant unless enforced, `null` usage never reads as exhausted).
- `client-entitlements.test.ts` proves the browser projection never carries a model name, and that it is an allowlist rather than a `delete`.
- `tab-feature.test.ts` covers the tab→feature map.
- Plans surfaces were exercised by hand against a local gateway via `.context/plans-ui/plans.sh`.

**No automated test yet** for the tools, the card, the paywall, the four route-level gates, or `resolveTier`'s pin branch. Those are e2e-shaped (real server + a gateway fake) and `packages/e2e` has no gateway fake yet — worth a follow-up.

**Not verified by hand:** Monitor below Ultra (needs seeded LLM stats — no OpenRouter credentials locally, so the cards read empty either way), the Billing states after a downgrade, and the plan card's loading/error/unavailable/warn states.


pedrofrxncx:
Reopened and fast-forwarded to `119db179` — the head of #7218, which is now closed.

**Why this PR and not that one:** #7218 was opened as a reopen of this branch after it was closed unmerged, but the work originated here. Rather than carry the review history and the authorship onto a second PR, the ten `apps/web` design commits were folded back onto `feat/tiered-plans-feature-gates`, so there is one PR, one branch, and one review thread again. `git merge-base --is-ancestor` confirmed #7218 was a strict descendant of this branch — nothing was dropped and no merge commit was needed.

**What came across:** `main` merged in (this branch was ~90 commits behind), plus the design pass — the single tier ladder (`plan-ladder.tsx`), the shared `PaywallDialog`/`FeaturePaywall`, sidebar locks via `use-tab-locked.ts`, the deduplicated `TopUpAmounts` picker under a new `credits` i18n domain, and the Free-org fallback in the two credits dialogs that previously `return null`-ed into a dead end.

**Authorship:** the design commits keep @rafavalls as their author in git. The body has been rewritten to describe both halves, and the gaps that came with the design pass (no paywall backdrop for CMS or Monitor, BYO keys orphaned on downgrade, the Stripe dev string in the subscribe toast, hardcoded `planPriceBrl`) are now listed alongside the server-side ones rather than living only on the closed PR.

Still blocked on decocms/ai-gateway#30 merging first.

pedrofrxncx:

[studio-bot:stale-close]

PR #7149, opened 2026-09-10
feat(repositories): search all installed repos server-side in the grant picker

guitavano:
## Summary

The GitHub connect **grant picker** ("Choose up to 500 repositories for this workspace") only filtered the repositories already paged into the client. A repo on a not-yet-loaded page was invisible to the search box, and the empty state misleadingly said there were no repositories.

This threads the query to the backend so the search runs server-side across every installed repository — matching the pattern the `REPOSITORY_SEARCH` picker already uses.

## Changes

**Backend**
- `app-auth.ts`: `listRepositoryChoices` takes an optional `query`, filtering `owner/name` per page via the existing (already-tested) `matchesRepoQuery`. `hasMore` stays on the raw page length, so the filter never hides later provider pages. `canAuthorizeRepositories` still calls it without a query (no regression).
- `git-providers.ts`: `GET .../flows/:id/repositories` accepts an optional `query` (trimmed, max 256) and passes it through.

**Frontend (`github-connect-dialog.tsx`)**
- Debounces the query (300ms), threads it into the fetch URL + query key, and drops the client-side `.filter()`.
- Empty state now distinguishes "load more to keep searching" / "no matches" / "no repositories".
- Placeholder copy: "Filter loaded repositories" → "Search repositories" (en + pt-br).

## Known limitation

GitHub's installation-repositories endpoint has no native text search, so results are still filtered per page — a match deep in the pages may need "Load more". The fix is that search now works *across* pages instead of only over what was already loaded.

## Testing
- `tsc --noEmit` clean in `apps/api` and `apps/web`.
- `bun run lint`: 0 errors, no warnings in changed files.
- `bun run fmt` applied.
- The pure `matchesRepoQuery` semantics this relies on are fully covered in `client.test.ts` (substring, case-insensitive, empty/blank → match-all). The route/hook wiring is e2e territory (needs a live GitHub App) with no existing coverage.

---
## Summary by cubic
The GitHub connect grant picker now searches all installed repositories server-side, so a repo on a page the client hasn't loaded yet shows up in search instead of appearing missing.

- `listRepositoryChoices` and the `/repositories` route take an optional `query` filtered with the existing `matchesRepoQuery`; `hasMore` still tracks the raw page length so the filter never hides later provider pages.
- The picker debounces the query 300ms, includes it in the fetch URL and query key, and drops the client-side filter.
- The empty state now distinguishes "no repositories", "no matches", and "load more to keep searching"; the placeholder is now "Search repositories" in en and pt-br.
- GitHub's installation endpoint has no native search, so matching stays per page and a deep match may still need "Load more".

<sup>Written for commit 2eb1fec52b7adbbe7581a11063e267042996f256. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7149?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/api/routes/git-providers.ts
+++ b/apps/api/src/api/routes/git-providers.ts
@@ -193,6 +193,7 @@ export const createGitProviderRoutes = () => {
       .object({
         installationId: z.coerce.number().int().positive().safe(),
         page: z.coerce.number().int().positive().safe().default(1),
+        query: z.string().trim().max(256).optional(),
       })
       .safeParse(c.req.query());
     if (!input.success) return c.json({ error: "Invalid installation" }, 400);
@@ -222,6 +223,7 @@ export const createGitProviderRoutes = () => {
         userToken,
         installation,
         input.data.page,
+        input.data.query,
       );
       const account = await ctx.storage.gitProviderAccounts.findByExternalId({
         organizationId: owner.organizationId,
--- a/apps/api/src/git-providers/github/app-auth.ts
+++ b/apps/api/src/git-providers/github/app-auth.ts
@@ -22,6 +22,7 @@ import { APP_BUDGET_OWNER, rememberBudgetOwner } from "./budget-owner";
 import { type GithubAppConfig, readGithubAppConfig } from "./env";
 import type { GitProviderCapability } from "../types";
 import { GitProviderError } from "../types";
+import { matchesRepoQuery } from "./client";
 import {
   githubApiBaseUrl,
   githubErrorMessage,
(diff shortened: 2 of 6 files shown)


viktormarinho (approved):
lgtm

PR #7218, opened 2026-09-15
feat(plans): tiered plans, feature gates and the AI usage bar

rafavalls:
The studio half of tiered plans, reopened. Supersedes #7116, which was closed unmerged — this branch is a strict superset of that PR's head (0 commits missing) plus `main` merged in and a design pass over every plans surface.

**Gateway half first:** `decocms/ai-gateway#30`. Without it `DECO_AI_GATEWAY_ENABLED` is unset, no gateway adapter exists, and every gate here fails open — so this PR is inert rather than broken.

## Authorship

The server-side work — `plan-feature-gate.ts`, `stripe-webhook.ts`, `entitlements.ts`, `topup-url.ts`, the `requiresFeature` wrapper on `defineTool` — is @pedrofrxncx's, carried over from #7116 unchanged. The commits on top (10, all `apps/web` plus four lines of `packages/ui/src/styles/global.css`) are the design pass.

## The split this PR preserves

**Studio stays open source and holds no pricing logic.** It hardcodes the feature *keys* — `cms`, `chat`, `kanban`, `monitoring`, `model_choice`, `diagnostic`, `diagnostic_enriched` — and every *value* is the gateway's answer for that org at that moment. No allowance, price or margin appears in this diff, with one marked exception below.

## The design pass on top

- **One tier ladder.** `plan-ladder.tsx` is the single answer to "what does this tier look like" — its mark, its colour, its price. The plan card, the catalog and the paywall all read from it, so they cannot disagree.
- **One paywall dialog.** `PaywallDialog` is the shell; `FeaturePaywall` fills it in. The gated panel renders a populated, inert board behind it so the dialog is not sitting on an empty screen. Site Editor and Monitor deliberately have no backdrop — see Known gaps.
- **Locks in the sidebar.** A row whose view the plan withholds carries a lock, defined once in `use-tab-locked.ts` so both nav lists agree.
- **One top-up picker.** The zero-balance and exhausted-mid-stream dialogs were byte-for-byte copies; `TopUpAmounts` is the shared one, under a new `credits` i18n domain.
- **A Free org is told why.** Both credits dialogs bailed with `return null` on a plan without `credits`. The empty state had no fallback, so an org whose allowance was spent opened a chat it could not run and was told nothing. It now shows the plan dialog.

## Needs a decision before merge

- **Prices are hardcoded** in `planPriceBrl`, keyed by plan id, until Stripe is wired. This is the one place a number appears in the diff, and it is marked as temporary. Keyed by id and not by rung on purpose: a mark on the wrong rung is cosmetic, a price on the wrong plan is a misquote.
- **`Pro+` wants to be `Scale`.** Product decision made; needs a gateway migration (`drizzle/0009_add_plans.sql` seeds the name), so it is not in this diff.
- **Below Ultra, gated UI disappears silently** — the tier pill in the composer, Default models and the BYO tiles on Billing, the Cost card on Monitor. Consistent with itself, but it contradicts the sidebar, which locks rather than hides. Worth settling in one pass across all three.

## Known gaps

- **No paywall backdrop for CMS or Monitor.** Kanban's works because `TaskBoardPage` reads one plain query key. Site Editor is a live sandbox iframe with nothing to seed; the CDN tab depends on a suspense query keyed by a runtime MCP client object plus ~7 derived keys and default component state. Left blank rather than coupled to two components' internals.
- **A downgrade orphans BYO provider keys.** `ConnectedProvidersSection` is gated on `model_choice`, so an org that drops from Ultra keeps its keys in the database with no UI to see or remove them. Not fixed here.
- **The subscribe error toast leaks a dev string** (`Stripe API 503: plan 'pro' has no Stripe price configured`).

## Not verified

Monitor below Ultra (needs seeded LLM stats — no OpenRouter credentials locally, so the cards read empty either way), the Billing states after a downgrade, and the plan card's loading/error/unavailable/warn states.

## Testing

`bun run check`, `bun run lint` and `bun run fmt` pass. Plans surfaces were exercised by hand against a local gateway via `.context/plans-ui/plans.sh`.

---
## Summary by cubic
Adds tiered plans with per-org feature gates and an AI usage bar, enforced server-side so a gated tool is covered by declaring `requiresFeature` once. Studio hardcodes only the feature keys; every value is the gateway's answer for that org, and every gate fails open without an answer — the PR is inert until `decocms/ai-gateway#30` lands.

**Behavior changes**
- Gated tools refuse at the execute wrapper; chat, monitoring and CMS gate at their own chokepoints — the messages POST, `requireOwnedSite`, and `resolveDecofileScope`.
- Stripe is the only way to acquire a paid tier: a price mapped in `STRIPE_PLAN_PRICE_IDS` grants the tier on an active subscription and revokes it on cancellation.
- Studio refuses to boot with plans on and no explicit gateway JWT secret or provision key, instead of silently opening every gate.
- The Kanban gate and budget stop apply at the task-dispatch paths and the two sandbox routes that spend, not just the board tools.
- BYO credentials and model pins are gated on `model_choice`; below Ultra the model's name and per-message costs are withheld, a gate that fails closed.
- The usage bar is a percent, never a dollar; it advances optimistically by each turn's own cost.

**Client surfaces**
- One tier ladder (`plan-ladder.tsx`) drives the plan card, catalog and paywall, so they cannot disagree; prices are hardcoded in `planPriceBrl` keyed by plan id until Stripe is wired.
- A sidebar row whose view the plan withholds carries a lock, defined once in `use-tab-locked` so both nav lists agree.
- The two credits dialogs share one top-up picker, and a Free org with a spent allowance now sees the plan dialog instead of nothing.
- Entitlements are not persisted to localStorage, so a stale cached decision can't paint paywalls or model names.
- Known gaps: only Kanban has a paywall backdrop, a downgrade orphans BYO keys with no UI to remove them, and `Pro+` will be renamed `Scale` in a separate gateway migration.

<sup>Written for commit 119db1793f48aacc9faa54b0cd0349c941134abe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7218?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:
Folded into #7116 — that branch now points at `119db179`, this PR's exact head. Closing in favour of it.

PR #7245, opened 2026-09-15
fix(ai-providers): scope the model-list cache to the key, not just the provider

pedrofrxncx:
A correctness bug found while auditing `apps/api/src/ai-providers/factory.ts` for stale-cache issues.

The `openai-compatible` adapter's `listModels()` hits the key's own custom `baseUrl` (`apps/api/src/ai-providers/adapters/openai-compatible.ts`), so two different keys on the same provider can have completely different catalogs. But `ModelListCache` keyed entries only by `organizationId` + `providerId` (`apps/api/src/ai-providers/model-list-cache.ts`), with no `keyId` dimension — the migration `041-ai-provider-keys-unique-constraint.ts` confirms an org can hold multiple distinct `openai-compatible` keys.

**Failure scenario:** an org creates openai-compatible key A pointing at `https://a.example.com`, calls `AI_PROVIDER_KEY_LIST_MODELS` (caching under `org.openai-compatible`), then creates key B pointing at a different self-hosted endpoint `https://b.example.com` and calls list-models again within the 10-minute TTL. Because the cache key ignores `keyId`, key B's request returns key A's cached model catalog — the wrong endpoint's models are shown for the wrong connection.

**Fix:** add `keyId` to `ModelListCache.get`/`set`/`invalidate` and thread it through `AIProviderFactory.listModels` (the OpenRouter enrichment index keeps using its existing global sentinel id, since it's provider-wide, not key-specific).

**Regression test:** `apps/api/src/ai-providers/factory.test.ts` — "caches openai-compatible model lists per key, not per provider" — two openai-compatible keys with different baseUrls must each see their own model, not a cross-key cached result. Verified the test fails on the pre-fix code (returns key A's model for key B) and passes after the fix.

Reviewer command: `bun test apps/api/src/ai-providers/factory.test.ts`

Locally ran: `bun run fmt`, `cd apps/api && bunx tsc --noEmit` (clean), `bunx oxlint` on the three touched files (clean), and the targeted test file above. Full CI validates the rest.

---
## Summary by cubic
Fixes a stale-cache bug in `openai-compatible` model lists: the cache was keyed by org and provider only, even though each key can have its own `baseUrl`, so a second key on the same org with a different `baseUrl` could see the first key’s cached models for up to the 10-minute TTL. Cache entries now include `keyId`, and a regression test verifies that two keys on the same org with different endpoints get separate model lists.

<sup>Written for commit db2aafef6779799c4c7b58f443cc419b41b09b9c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7245?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:

[studio-bot:stale-close]

PR #7278, opened 2026-09-16
fix(task-board): don't show the cross-org banner for a rejected boardOrg

pedrofrxncx:
Follows #7276 (cross-org admin task board picker), which added `?boardOrg=` to view another tenant's board from your own org.

`BoardOrgProvider` (board-org.tsx) fails closed: when `?boardOrg=<slug>` names an org the server didn't list — the caller isn't an admin-org member, or the slug is bogus/stale — it silently renders the caller's own board instead of the requested one. `TaskBoardAdminBanner`, however, rendered purely off the raw `?boardOrg=` search param (`useBoardOrgSlug()`), with no check that the swap actually applied. So any user who ends up with an invalid `?boardOrg=<slug>` in the URL (a stale bookmark, a shared link, or simply losing admin-org access after bookmarking one) sees a banner claiming "You're viewing `<slug>`'s board" while the board underneath is actually their own org's data — a misleading claim about data provenance on a cross-tenant surface.

Fix: added `useBoardOrgTarget()` in board-org.tsx, which resolves `?boardOrg=` against the server-returned admin org list and returns `null` unless the target is confirmed valid — the same check `BoardOrgProvider` already applies. `TaskBoardAdminBanner` now renders off this validated target instead of the raw slug, so it can never disagree with what the board is actually showing.

Failure scenario: sign in as a non-admin (or an admin-org member no longer in an admin org) and load `/{org}/tasks?boardOrg=some-other-org`. Before this fix: the banner says "You're viewing some-other-org's board" while the board shown is your own org's. After: no banner, matching the board actually rendered.

Reviewer check: `cd apps/web && bunx tsc --noEmit` (clean on the touched files) and read board-org.tsx / admin-controls.tsx — `useBoardOrgTarget()` is the single source both the provider and the banner key off.

Locally ran: `bun run fmt`, `bunx tsc --noEmit` (apps/web), `bunx oxlint` on both changed files — all clean. No existing test covers this component pair; full CI validates the rest.

---
## Summary by cubic
Prevents the task board banner from claiming users are viewing another organization’s board when `?boardOrg=` is invalid or rejected. The banner now appears only when the requested organization is confirmed by the server and the board switch is applied.

- Shares the validated target lookup between `BoardOrgProvider` and `TaskBoardAdminBanner`.
- Invalid, stale, or unauthorized targets fail closed without showing the banner.

<sup>Written for commit a4ffe1cbb05dadd707f2432c3e7af7af017abdb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:

[studio-bot:stale-close]

PR #7303, opened 2026-09-17
fix(ui): collapse SearchToggle on Escape instead of only clearing it

pedrofrxncx:
**Bug:** in `packages/ui/src/components/search-toggle.tsx`, the Escape-key handler only called `onChange("")` — it never closed the toggle. Since the input was still focused when Escape fired, the "collapse when cleared from outside" check (`value === "" && !focused`) never triggered, so pressing Escape left the search field open and empty instead of dismissing it back to the icon button, which is the behavior every other search-toggle/combobox in the app already gives (Escape = dismiss).

**Fix:** when Escape is pressed and the toggle isn't in `expanded` mode, clear the value, `setOpen(false)`, and blur the input so it collapses back to the trigger button — matching the existing blur-when-empty behavior, just reachable from the keyboard.

**Verify:** open any toolbar using `SearchToggle` (e.g. a task-board or library toolbar search), type something, press Escape — the field now clears and collapses back to the icon button instead of staying open empty.

**Checked locally:** `bun run fmt`, `cd packages/ui && bunx tsc --noEmit` (no new errors — the file itself is clean, pre-existing `.stories.tsx` TS4023 noise is unrelated), `bunx oxlint packages/ui/src/components/search-toggle.tsx` (0 warnings/errors). No render-test harness exists yet for `packages/ui` components (Playwright CT lives under `apps/web/ct`), so this was verified by tracing the state logic; full CI covers the rest.

---
## Summary by cubic
Fixes the Escape key in `SearchToggle` so it dismisses the search field instead of just clearing the typed value. Previously, pressing Escape left the field open and empty; now it collapses back to the icon button, matching the behavior of other search toggles in the app. The `expanded` mode keeps its existing behavior.

<sup>Written for commit eaeac99529b2f28660a3b6f8e2d337b8cfcc54c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


pedrofrxncx:

[studio-bot:stale-close]

PR #7316, opened 2026-09-17
feat(jira): start a run on several issues at once

viktormarinho:
First half of the manual-fire work. The merge action lands next, on the same field component.

## Why

Firing runs one at a time is how the manual surface is actually used: you work a batch off a board, paste a column of links, and repeat the same click ten times.

## What

`JIRA_RUN_START` takes several issues — keys or links, newline / comma / semicolon separated — and starts one run each.

Two things it deliberately does not do quietly:

- **A piece it cannot read comes back in `unreadable`** rather than being dropped. Starting nine runs and saying nothing about the tenth is the failure this return shape exists to prevent.
- **One issue that fails to start does not fail the batch.** Each lands in `failed` with its reason, so the other nine still go and you know which one to look at.

Starts are **sequential**: each reads its issue from Jira and claims a run slot, so a paste of twenty shouldn't become a burst against either. Capped at 25 per call — each one is a real, paid run, and a fat-fingered paste of a whole board should be refused up front rather than discovered as a bill.

## The parser moved to `packages/shared`

`@decocms/shared/jira/issue-key`. The field now parses **as you type**, so the card reports "10 issues" *before* the button that starts ten runs, and names a key it could not read while there's still time to fix it. The parser is pure and isomorphic, which is what that package is for.

## UI

The issue field is now a component **both manual actions share** (`IssueKeysField`) — the merge card in the follow-up uses the same one, which is the consistency this was asked for.

The result is a **list on the card, not a toast**: ten issues produce ten outcomes, and the two that failed are the ones worth reading. It stays until the next run.

## Testing

- `parseIssueKeys`: mixed separators, a pasted column of links, unreadable pieces surfacing rather than dropping, repeats collapsing (firing an issue twice would just have the second supersede the first), empty input.
- `bun run check` / `fmt` / `lint` / `knip` clean. The one failing test in the touched tree is `trigger.integration.test.ts` on `ECONNREFUSED` — no local Postgres.

## Noticed while here

`JIRA_RUN_START` has **no `kind` parameter** — I had been passing `kind: "execute"` / `"review"` when firing batches by hand and Zod was silently stripping it. Harmless, because the execute/review distinction comes from the skill text baked into `prompt` (that's the design from #7184), but worth stating so nobody else assumes the field does something.

---
## Summary by cubic
Starts Jira runs for multiple issues at once, so a pasted column of links fires one run per issue instead of requiring a separate click each.

**Details**
- `JIRA_RUN_START` now takes multiple issue keys or links (separated by newlines, commas, or semicolons) and starts one run each, sequentially, capped at 25.
- The output reports `started`, `failed` (with reasons), and `unreadable` pieces; one bad issue doesn't fail the batch, and unreadable input isn't silently dropped.
- The issue-key parser moved to `@decocms/shared/jira/issue-key`; the settings UI now shows the parsed count and unreadable keys as you type, and batch results appear as a list on the card instead of a toast.
- Note: `JIRA_RUN_START` has no `kind` parameter; the execute/review distinction comes from the prompt text.

<sup>Written for commit d528389a787febe77779656e8336df9e7f8cb829. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7316?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


viktormarinho:
Folded into #7317, which carries both commits — one PR, as asked.

PR #7379, opened 2026-09-18
fix(task-board): duplicate check survives a reasoning model's answer

viktormarinho:
## Summary

Second production follow-up to #7365 (#7376 made the outcome visible; this fixes what it revealed).

With `duplicateCheck` in the output, the throwaway-org test now showed `skipped: unparseable answer from qwen/qwen3.5-flash-02-23` on two of three calls. The org's fast tier resolves to a reasoning model: it thinks before it answers, so the JSON arrives after a `<think>` block, after narration that may itself contain braces, or not at all within the 1200-token budget.

## Changes

- `parseModelJson` strips `<think>…</think>`, ignores code fences anywhere, and tries every top-level balanced `{…}` from the **last** one backwards, so narration before the answer never wins.
- `maxOutputTokens` 1200 → 8000, room for the model to reason before the JSON.
- The skip reason now carries `finishReason` and the answer length, so a truncation reads as `finish=length` instead of a generic "unparseable". The warn log gets the tail of the answer rather than the head.
- Two parser tests: a `<think>` block with braces inside, and narration with an earlier JSON object before the real one.

## Testing

28 tests in `duplicate-check.test.ts`, `bun run check`, `bun run lint`, `bun run knip`, `bun run fmt`. Will re-run the production scenario on the same org once deployed and report on this PR.

---
## Summary by cubic
Fixes task-board duplicate check for reasoning models so their JSON answer is used instead of being skipped as unparseable.

**Bug Fixes**
- `parseModelJson` strips `thinking` blocks, ignores code fences, and tries top-level JSON objects from last to first.
- Raises the fast tier `maxOutputTokens` from 1200 to 8000, which may increase cost/latency when the extra budget is used.
- Skip reasons now include the finish reason and answer length so truncation is distinguishable from malformed output.
- Adds parser tests for a reasoning block with braces and narration containing an earlier JSON object.

<sup>Written for commit b5e621d2818d6e78fa44cb232b21a0d4b85c9e09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/decocms/studio/pull/7379?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


viktormarinho:
Production data behind this PR, from the throwaway org:

- With the org's default fast pick (a qwen flash reasoning model): 3/3 `skipped: unparseable answer` once the board had a dozen cards. Longer boards mean longer thinking, so the JSON never fits in 1200 tokens or arrives after a `<think>` block.
- With the org's fast tier pinned to `anthropic/claude-haiku-4.5` via `ORGANIZATION_SETTINGS_UPDATE`, on the code already deployed (#7376, before this PR): 3/3 `matched` on paraphrases (EN↔PT, both directions), 2/2 `no_match` on related-but-distinct tasks (same area, different change).

So the judgement and the confidence gate are right; the remaining failure is purely the answer format of reasoning models on the default pick, which is what this PR addresses. Orgs that pin a non-reasoning fast model already work without it.

